PR #27, opened 2023-04-17
Upload file endpoint implementation

### LINKED ISSUE #13: Upload file
Given uri:
`https://api.openai.com/v1/files`

Create sttp POST endpoint + tests

Docs reference
https://platform.openai.com/docs/api-reference/files/upload

benzwreck:
Closes #13

### DIFF
--- a/core/src/main/scala/sttp/openai/OpenAi.scala
+++ b/core/src/main/scala/sttp/openai/OpenAi.scala
@@ -6,6 +6,8 @@ import sttp.openai.requests.models.ModelsGetResponseData.ModelsResponse
 import sttp.openai.json.SttpUpickleApiExtension.asJsonSnake
 import sttp.openai.requests.files.FilesResponseData._
 
+import java.io.File
+
 class OpenAi(authToken: String) {
 
   /** Fetches all available models from [[https://platform.openai.com/docs/api-reference/models]] */
@@ -19,6 +21,27 @@ class OpenAi(authToken: String) {
     openApiAuthRequest
       .get(OpenAIEndpoints.FilesEndpoint)
       .response(asJsonSnake[FilesResponse])
+
+  /** Upload a file that contains document(s) to be used across various endpoints/features. Currently, the size of all the files uploaded by
+    * one organization can be up to 1 GB. Please contact OpenAI if you need to increase the storage limit.
+    * @param file
+    *   [[java.io.File File]] of the JSON Lines file to be uploaded. <p> If the purpose is set to "fine-tune", each line is a JSON record
+    *   with "prompt" and "completion" fields representing your
+    *   [[https://platform.openai.com/docs/guides/fine-tuning/prepare-training-data training examples]].
+    * @param purpose
+    *   The intended purpose of the uploaded documents. <p> Use "fine-tune" for Fine-tuning. This allows OpenAI to validate the format of
+    *   the uploaded file.
+    * @return
+    *   Uploaded file's basic information.
+    */
+  def uploadFile(file: File, purpose: String = "fine-tune"): Request[Either[ResponseException[String, Exception], FileData]] =
+    openApiAuthRequest
+      .post(OpenAIEndpoints.FilesEndpoint)
+      .multipartBody(
+        multipart("purpose", purpose),
+        multipartFile("file", file)
+      )
+      .response(asJsonSnake[FileData])
 
   private val openApiAuthRequest: PartialRequest[Either[String, String]] = basicRequest.auth
     .bearer(authToken)
--- a/core/src/main/scala/sttp/openai/OpenAi.scala
+++ b/core/src/main/scala/sttp/openai/OpenAi.scala
@@ -7,6 +7,7 @@ import sttp.openai.json.SttpUpickleApiExtension.asJsonSnake
 import sttp.openai.requests.files.FilesResponseData._
 
 import java.io.File
+import java.nio.file.Paths
 
 class OpenAi(authToken: String) {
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -20,6 +22,27 @@ class OpenAi(authToken: String) {
...
+    *   [[java.io.File File]] of the JSON Lines file to be uploaded. <p> If the purpose is set to "fine-tune", each line is a JSON record
+    *   with "prompt" and "completion" fields representing your
+    *   [[https://platform.openai.com/docs/guides/fine-tuning/prepare-training-data training examples]].
+    * @param purpose
+    *   The intended purpose of the uploaded documents. <p> Use "fine-tune" for Fine-tuning. This allows OpenAI to validate the format of
+    *   the uploaded file.
+    * @return
+    *   Uploaded file's basic information.
+    */
+  def uploadFile(file: File, purpose: String = "fine-tune"): Request[Either[ResponseException[String, Exception], FileData]] =
```
flsh86, line 37: `suggestion` How about we overload this method with another one that accept String as an argument, that is a path to a file?
@kciesielski @adamw

flsh86 (approved):

core/src/main/scala/sttp/openai/OpenAi.scala
kciesielski: I think it might be useful, similarly to Tapir where we have endpoints that take a `systemPath: String`.

benzwreck: Okay, so here's the deal. [We cannot overload a method with default parameters](https://stackoverflow.com/questions/4652095/why-does-the-scala-compiler-disallow-overloaded-methods-with-default-arguments). So from my point of view we could either get rid of `def uploadFile(file: File, purpose: String = "fine-tune")` and leave `def uploadFile(systemPath: String, purpose: String = "fine-tune")` alone or create four methods e.g.
```
def uploadFile(file: File, purpose: String)
def uploadFile(file: File) // with hard-coded 'purpose' as 'fine-tune'
def uploadFile(systemPath: String, purpose: String)
def uploadFile(systemPath: String) // with hard-coded 'purpose' as 'fine-tune'
```
that will handle all possible options. Wdyt?

adamw: yes that's the way to do it :)

benzwreck: @kciesielski @adamw could you please take a look?

kciesielski: @benzwreck +1 to Adam, use multiple method variants instead of default parameters.

core/src/test/scala/sttp/openai/requests/files/FilesResponseDataSpec.scala
```diff
@@ -34,4 +34,26 @@ class FilesResponseDataSpec extends AnyFlatSpec with Matchers with EitherValues
...
+        purpose = "fine-tune",
+        filename = "example.jsonl",
+        bytes = 44,
+        createdAt = 1681375533,
+        status = "uploaded",
+        statusDetails = None
+      )
+
+    // when
+    val givenResponse: Either[Exception, FileData] = SttpUpickleApiExtension.deserializeJsonSnake[FileData].apply(singleFileJsonResponse)
```
kciesielski, line 54: **question:** Is `Exception` the strongest error we can get here?

kciesielski: Ok, I checked upickle sources and the API indeed can throw a few kinds of `Exception`, some of them upickle-specific, sometimes raw Java exceptions like `IllegalArgumentException`, which is not nice.
However, I noticed in SnakePickle, that you're handling errors there with:
```scala
    try
      Right(SnakePickle.read[B](JsonInput.sanitize[B].apply(s)))
    catch {
      case e: Exception => Left(e)
      case t: Throwable =>
        // in ScalaJS, ArrayIndexOutOfBoundsException exceptions are wrapped in org.scalajs.linker.runtime.UndefinedBehaviorError
        t.getCause match {
          case e: ArrayIndexOutOfBoundsException => Left(e)
          case _                                 => throw t
        }
    }
```
How about changing both occurrences of `Left(e)` to `Left(new DeserializationException(e)`, where `DesrializationException` will be your own type, defined in `sttp.openai.json`? This could allow slightly better assertions here, wdyt?

core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -80,6 +67,103 @@ class OpenAi(authToken: String) {
...
+  /** Fetches all files that belong to the user's organization from [[https://platform.openai.com/docs/api-reference/files]] */
+  def getFiles: Request[Either[ResponseException[String, Exception], FilesResponse]] =
+    openApiAuthRequest
+      .get(OpenAIEndpoints.FilesEndpoint)
+      .response(asJsonSnake[FilesResponse])
+
+  /** Upload a file that contains document(s) to be used across various endpoints/features. Currently, the size of all the files uploaded by
+    * one organization can be up to 1 GB. Please contact OpenAI if you need to increase the storage limit.
+    * @param file
+    *   [[java.io.File File]] of the JSON Lines file to be uploaded. <p> If the purpose is set to "fine-tune", each line is a JSON record
```
adamw, line 79: [minor] what's the documentation benefit of `[[java.io.File File]]`? The parameter type is in the signature already, as well as the name is known?

PR #35, opened 2023-04-19
Create image

flsh86:


benzwreck (approved):

core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -95,11 +108,17 @@ class OpenAi(authToken: String) {
 }
 
 private object OpenAIEndpoints {
+  private val ImageEndpointBase: Uri = uri"https://api.openai.com/v1/images"
```
kciesielski, line 111: :ok_hand:

core/src/main/scala/sttp/openai/requests/images/ImageCreationRequestBody.scala
```diff
@@ -0,0 +1,96 @@
...
+  sealed abstract class Size(val value: String)
+
+  object Size {
+    case object Small extends Size("256x256")
+
+    case object Medium extends Size("512x512")
+
+    case object Large extends Size("1024x1024")
+
+    case class Custom(customSize: String) extends Size(customSize)
```
kciesielski, line 64: **suggestion:** Please add a Scaladoc comment like "Use only as a workaround if API supports a format that's not yet predefined as a case object of Size. Otherwise, a custom format would be rejected. See (link here) for current list of supported formats."

core/src/main/scala/sttp/openai/requests/images/ImageCreationRequestBody.scala
```diff
@@ -0,0 +1,96 @@
...
+      .readwriter[ujson.Value]
+      .bimap[Size](
+        _.value,
+        jsonValue =>
+          SnakePickle.read[ujson.Value](jsonValue) match {
+            case Str(value) =>
+              withValue(value) match {
+                case Some(size) => size
+                case None       => NotSupportedSize
+                // TODO
```
kciesielski, line 85: **suggestion:** I think adding a NotSupportedSize type may be misleading. The `Size` type will be used almost exclusively for requests and we don't want to confuse users with giving them such an option. Maybe we can deserialize to a `Custom` format here?

kciesielski (approved):

PR #37, opened 2023-04-20
List fine tunes endpoint implementation

benzwreck:
Should be merged after #34

### DIFF
--- a/core/src/main/scala/sttp/openai/Main.scala
+++ b/core/src/main/scala/sttp/openai/Main.scala
@@ -1,16 +1,13 @@
 package sttp.openai
 
 import sttp.client4._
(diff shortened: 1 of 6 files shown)


kciesielski (approved):

core/src/main/scala/sttp/openai/Main.scala
```diff
@@ -6,8 +6,9 @@ object Main extends App {
   val backend: SyncBackend = DefaultSyncBackend()
 
   val openAi: OpenAi = new OpenAi("sk-")
```
flsh86, line 8: Let's maybe not modify the main, idealiy would be to clear it or leave it as it was?

PR #41, opened 2023-04-21
Implement create embeddings endpoint

### LINKED ISSUE #40: Create embeddings
Given uri:
https://api.openai.com/v1/embeddings

Create sttp POST endpoint + tests

Docs reference
https://platform.openai.com/docs/api-reference/embeddings

benzwreck:
Closes #40


core/src/main/scala/sttp/openai/requests/embeddings/EmbeddingsResponseBody.scala
```diff
@@ -0,0 +1,33 @@
...
+import sttp.openai.json.SnakePickle
+
+object EmbeddingsResponseBody {
+  case class EmbeddingData(
+      `object`: String,
+      index: Int,
+      embedding: Seq[Double]
+  )
+  object EmbeddingData {
+    implicit val embeddingDataReader: SnakePickle.Reader[EmbeddingData] = SnakePickle.macroR[EmbeddingData]
```
adamw, line 12: [minor] Maybe it would be enough to call `SnakePickle.macroR` (no type parameter - it's inferred)
[question] is `macroR` a upickle-defined method? the name is the quite poor ;) (it should be `derived` or sth like that, it's not interesting that it's a macro)

benzwreck: `[minor] Maybe it would be enough to call SnakePickle.macroR (no type parameter - it's inferred)`
I guess not, compiler complains:
```
[error] 12 |    implicit val embeddingDataReader = SnakePickle.macroR[EmbeddingData]
[error]    |                 ^
[error]    |                type of implicit definition needs to be given explicitly
```
`[question] is macroR a upickle-defined method? the name is the quite poor ;) (it should be derived or sth like that, it's not interesting that it's a macro)`
Yes, it is upickle-defined method :)

adamw: > I guess not, compiler complains:

Yeah the type annotation must remain, I meant the type parameter to `macroR`

benzwreck: Oh, you are right. Got rid of them in `embeddings` module for now and will do a clean-up in others in next PR

PR #50, opened 2023-04-25
Implement delete fine-tune model endpoint

### LINKED ISSUE #21: Delete fine-tune model
Given uri:
`https://api.openai.com/v1/models/{model}`

Create sttp DELETE endpoint + tests

Docs reference
https://platform.openai.com/docs/api-reference/fine-tunes/delete-model

benzwreck:
Closes #21

### DIFF
--- a/core/src/main/scala/sttp/openai/OpenAi.scala
+++ b/core/src/main/scala/sttp/openai/OpenAi.scala
@@ -13,7 +13,7 @@ import sttp.openai.requests.embeddings.EmbeddingsRequestBody.EmbeddingsBody
 import sttp.openai.requests.embeddings.EmbeddingsResponseBody.EmbeddingResponse
 import sttp.openai.requests.files.FilesResponseData._
 import sttp.openai.requests.finetunes.FineTunesRequestBody
-import sttp.openai.requests.finetunes.FineTunesResponseData.{FineTuneResponse, GetFineTunesResponse}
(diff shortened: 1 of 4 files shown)


core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -475,6 +487,7 @@ private object OpenAIEndpoints {
   val VariationsImageEndpoint: Uri = ImageEndpointBase.addPath("variations")
 
   def deleteFileEndpoint(fileId: String): Uri = FilesEndpoint.addPath(fileId)
+  def deleteFineTuneModel(model: String): Uri = ModelEndpoint.addPath(model)
```
kciesielski, line 490: **suggestion:** `ModelEndpoint.addPath(model)` doesn't imply that it's a "delete" endpoint, only that it's a "modelEndpoint". Actually, similar remark applies to other methods here.

kciesielski (approved):

adamw (approved):

core/src/main/scala/sttp/openai/OpenAi.scala
benzwreck: it seems like it is not an endpoint but more like an URI so shall we rename it to `deleteFineTuneModelUri` then?

adamw: yeah this sounds good :)

flsh86 (approved):

PR #51, opened 2023-04-25
Implement retrieve file content endpoint

### LINKED ISSUE #15: Retrieve file content
Given uri:
`https://api.openai.com/v1/files/{file_id}/content`

Create sttp GET endpoint + tests

Docs reference
https://platform.openai.com/docs/api-reference/files/retrieve-content

benzwreck:
Closes #15

### DIFF
--- a/core/src/main/scala/sttp/openai/OpenAi.scala
+++ b/core/src/main/scala/sttp/openai/OpenAi.scala
@@ -39,7 +39,7 @@ class OpenAi(authToken: String) {
   /** Fetches all available models from [[https://platform.openai.com/docs/api-reference/models]] */
   def getModels: Request[Either[ResponseException[String, Exception], ModelsResponse]] =
     openApiAuthRequest
-      .get(OpenAIUris.Models)
+      .get(OpenAiUris.Models)
       .response(asJsonSnake[ModelsResponse])
 
   /** @param modelId
@@ -49,7 +49,7 @@ class OpenAi(authToken: String) {
     */
   def retrieveModel(modelId: String): Request[Either[ResponseException[String, Exception], ModelData]] =
     openApiAuthRequest
-      .get(OpenAIUris.model(modelId))
+      .get(OpenAiUris.model(modelId))
       .response(asJsonSnake[ModelData])
 
   /** @param completionBody
@@ -60,7 +60,7 @@ class OpenAi(authToken: String) {
     */
   def createCompletion(completionBody: CompletionsBody): Request[Either[ResponseException[String, Exception], CompletionsResponse]] =
     openApiAuthRequest
-      .post(OpenAIUris.Completions)
+      .post(OpenAiUris.Completions)
       .body(completionBody)
       .response(asJsonSnake[CompletionsResponse])
 
@@ -71,7 +71,7 @@ class OpenAi(authToken: String) {
     */
   def createImage(imageCreationBody: ImageCreationBody): Request[Either[ResponseException[String, Exception], ImageResponse]] =
     openApiAuthRequest
-      .post(OpenAIUris.CreateImage)
+      .post(OpenAiUris.CreateImage)
       .body(imageCreationBody)
       .response(asJsonSnake[ImageResponse])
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -475,6 +485,7 @@ private object OpenAIEndpoints {
   val VariationsImageEndpoint: Uri = ImageEndpointBase.addPath("variations")
 
   def deleteFileEndpoint(fileId: String): Uri = FilesEndpoint.addPath(fileId)
+  def retrieveFileContentEndpoint(fileId: String): Uri = FilesEndpoint.addPath(fileId, "content")
```
kciesielski, line 488: **suggestion:** Similar comment to what I've said on the delete method. Only the `.get()` call above determines that this is indeed a "retrieve" operation, the same endpoint may be used with over HTTP verbs for other purposes. Let's call this method just `fileContentEndpoint`.

adamw: ah, now I get it. I think the previous comment that this should be called an `...Uri` not an `...Endpoint` applies

benzwreck:
I've changed all occurrences of `...Endpoint` to `...Uri` but still have some doubts if we really need to explicitly include that in a variable/method name. Shouldn't type speak for itself? All of them are URIs and we refer to them by calling `OpenAIUris.cancelFineTune(id)` so how about something like that:
```scala
private object OpenAIUris {
  private val ImageBase: Uri = uri"https://api.openai.com/v1/images"
  private val AudioBase: Uri = uri"https://api.openai.com/v1/audio/"
  // ...
  val ChatCompletions: Uri = uri"https://api.openai.com/v1/chat/completions"
  val Completions: Uri = uri"https://api.openai.com/v1/completions"
  // ...
  def cancelFineTune(fineTuneId: String): Uri = FineTunes.addPath(fineTuneId, "cancel")
  def deleteFile(fileId: String): Uri = Files.addPath(fileId)
  def retrieveFileContent(fileId: String): Uri = Files.addPath(fileId, "content")
  // ...
```
@adamw @kciesielski @flsh86 wdyt?

adamw:
@benzwreck yeah that's true, it's in the object's name

kciesielski:
I meant something else. For example, you called a method `deleteFile` in
```scala
  def deleteFile(fileId: String): Uri = Files.addPath(fileId)
```
The fact that it's used for delete is not in the Uri, but in a DELETE call to this Uri. This is just a Uri which can have more purposes that deleting, when you call it with GET or POST or other HTTP method. That's why I suggest to name methods like this just `fileEndpoint`, or rather `file` if we agreed that endpoint is something else.

Note that there may be exceptions to this, like `cancelFineTune`, which indeed is a Uri meant only for canceling, because it has `/cancel` in the path.

core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -558,36 +568,36 @@ class OpenAi(authToken: String) {
...
-  val VariationsImageEndpoint: Uri = ImageEndpointBase.addPath("variations")
-
-  def cancelFineTuneEndpoint(fineTuneId: String): Uri = FineTunesEndpoint.addPath(fineTuneId, "cancel")
-  def deleteFileEndpoint(fileId: String): Uri = FilesEndpoint.addPath(fileId)
-  def deleteFineTuneModelUri(model: String): Uri = ModelEndpoint.addPath(model)
-  def listFineTunesEndpoint(fineTuneId: String): Uri = FineTunesEndpoint.addPath(fineTuneId, "events")
-  def retrieveFileEndpoint(fileId: String): Uri = FilesEndpoint.addPath(fileId)
-  def retrieveFineTuneEndpoint(fineTuneId: String): Uri = FineTunesEndpoint.addPath(fineTuneId)
-  def retrieveModelEndpoint(modelId: String): Uri = ModelEndpoint.addPath(modelId)
+private object OpenAIUris {
```
adamw, line 578: 👍 much better :)

adamw (approved):

kciesielski (approved):

PR #55, opened 2023-04-27
Edit missing fields

flsh86:


README.md
```diff
@@ -0,0 +1,112 @@
+![sttp-model](https://github.com/softwaremill/sttp-openai/raw/master/banner.png) Banner should be changed to sttp openAi
+
+[![Join the chat at https://gitter.im/softwaremill/sttp](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/softwaremill/sttp)
+[![CI](https://github.com/softwaremill/sttp-openai/workflows/CI/badge.svg)](https://github.com/softwaremill/sttp-openai/actions?query=workflow%3ACI+branch%3Amaster)
+
+[//]: # ([![Maven Central]&#40;https://maven-badges.herokuapp.com/maven-central/com.softwaremill.sttp&#41;&#40;https://maven-badges.herokuapp.com/maven-central/com.softwaremill.sttp&#41;)
```
adamw, line 6: what does `[//]: #` do here?

README.md
```diff
@@ -0,0 +1,112 @@
...
+
+sttp is a family of Scala HTTP-related projects, and currently includes:
```
adamw, line 8: Let's just write a short introduction, that sttp openai uses sttp client to describe requests to OpenAI endpotins or sth like that

README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Quickstart with sbt
+
+Add the following dependency:
+
+[//]: # (```scala)
+[//]: # ("com.softwaremill.sttp.openai" %% "core" % "?.?.?")
+[//]: # (```)
+
+sttp openai is available for Scala 2.12 and Scala 3.2.2
```
adamw, line 22: 2.13? ;)

README.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Project content
+
+Available endpoints with modeled response classes include:
```
adamw, line 26: Hm maybe it would be easier to point to OpenAI docs (to a specific version, if they are versioned) and list the endpoints which are not implemented?

adamw:
Sorry GH folded the class so I didn't notice ;)

README.md
```diff
@@ -0,0 +1,112 @@
...
+
+val bodyMessages: Seq[Message] = Seq(
+  Message(
+    role = "user",
+    content = "Hello!"
+  )
+)
+
+val chatRequestBody: ChatBody = ChatBody(
+  model = "gpt-3.5-turbo",
```
adamw, line 85: shouldn't this be an enum constant?

README.md
```diff
@@ -0,0 +1,112 @@
...
+  messages = bodyMessages
+)
+
+// Use createChatCompletion and pass created request body to create sttp request
+
+val request: Request[Either[ResponseException[String, Exception], ChatResponse]] = openAi.createChatCompletion(chatRequestBody)
```
adamw, line 91: let's maybe remove these type annotations, they look quite scary ;)

README.md
```diff
@@ -0,0 +1,112 @@
...
+// To invoke request and get a response provide your wished backend and send created request
+
+val backend: SyncBackend = DefaultSyncBackend()
+
+val response: Response[Either[ResponseException[String, Exception], ChatResponse]] = request.send(backend)
+```
+
+## Contributing
+
+If you have a question, or hit a problem, feel free to ask on our [gitter channel](https://gitter.im/softwaremill/sttp-model)!
```
adamw, line 102: No gitter anymore, sorry, sttp-model needs updating as well :)

core/src/main/scala/sttp/openai/OpenAi.scala
```diff
@@ -36,53 +36,56 @@ import java.nio.file.Paths
 
 class OpenAi(authToken: String) {
 
-  /** Fetches all available models from [[https://platform.openai.com/docs/api-reference/models]] */
+  /** Lists the currently available models, and provides basic information about each one such as the owner and availability.
+    * <p>[[https://platform.openai.com/docs/api-reference/models]]
```
adamw, line 40: is it fine to have an unclosed `<p>` tag?

README.md
flsh86: It comments it out, since nothing is posted yet under sttp.openai in maven repo it cannot load svg and link is pointing to non existing resource.

We can just remove it for now if you wish and add it once we have it posted on mvn

core/src/main/scala/sttp/openai/OpenAi.scala
benzwreck: according to javadocs - it is fine, but in scaladocs we can use blank line as a paragraph separator
`Paragraphs are started with one (or more) blank lines. * in the margin for the comment is valid (and should be included) but the line should be blank otherwise.`
https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html
so I'll change that <p> into blank lines then

README.md
flsh86: It can be, but then we have to change 
```
  case class ChatBody(
      model: String, ...)
```
To ADT smth like:
```
  case class ChatBody(
      model: ChatModel, ...)
```
According to docs there are different models to different functionality. 
https://platform.openai.com/docs/models/model-endpoint-compatibility

So we can either release it with String and make another release with missing models ADT, wdyt?

adamw: hm so the same `ChatBody` is used for different endpoints, and each endpoint has a limited subset of models?

flsh86: ChatBody is used only for createChatCompletion method.
Other methos that uses other type of models should have their own ADT. For example in Edits endpoint we create CompletionsBody
```
  case class CompletionsBody(
      model: String ...
```
uses string for now, but we can use ADT, that we should  explicity name as `EditsModel` because API accepts only one of models for Edits, which are text-davinci-edit-001 and code-davinci-edit-001 and those models are not shared with other endpoints/functionality

flsh86: I mean, without a doubt we should be using an ADT here as we are using in other parts of the project already. The question is, do we want to realease it first and update it in next release or postpone the relase and "fix" it first

adamw: Sure, let's create an issue an fix it later.

kciesielski (approved):

PR #72, opened 2023-05-05
Add a synchronous client

benzwreck:


README.md
```diff
@@ -88,7 +79,66 @@ println(response)
 )
 */
 ```
+#### Currently only two backend implementations are available:
+* `OpenAISyncBackend` which uses `Identity` as an effect `F[_]`
```
kciesielski, line 83: **nitpick:** Although `Id[A]` is called _identity monad_, here it implies that it's the type name, so maybe it should be 
```
which uses identity monad `Id[A]` as an effect `F[A]`
```

core/src/test/scala/sttp/openai/client/SyncClientSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+      |    "param": null,
+      |    "code": "invalid_api_key"
+      |  }
+      |}""".stripMargin
+
+  for ((statusCode, expectedError) <- testData)
+    s"Service response with status code: $statusCode" should s"return ${expectedError.getSimpleName}" in {
+      val syncBackendStub: SyncBackendStub = DefaultSyncBackend.stub.whenAnyRequest.thenRespondWithCode(statusCode, errorResponse)
+      val syncClient = OpenAISyncClient(authToken = "test-token", backend = syncBackendStub)
+      syncClient.getModels.left.value.getClass.getName shouldBe expectedError.getName
```
kciesielski, line 40: **suggestion:** You can compare just classes, without `getName`.

core/src/main/scala/sttp/openai/OpenAISyncClient.scala
```diff
@@ -0,0 +1,447 @@
...
+
+class OpenAISyncClient private (authToken: String, backend: SyncBackend) {
+
+  private val openAI = new OpenAI(authToken)
+
+  /** Lists the currently available models, and provides basic information about each one such as the owner and availability.
+    *
+    * [[https://platform.openai.com/docs/api-reference/models]]
+    */
+  def getModels: Either[OpenAIError, ModelsResponse] =
```
adamw, line 50: I think the sync client should use exceptions instead of `Either`s - providing a quick gateway to openai; if somebody wants more control over error cases, there's the `F`-backend which exposes eithers

core/src/main/scala/sttp/openai/OpenAISyncClient.scala
```diff
@@ -0,0 +1,447 @@
...
+          case Conflict                                     => TryAgain(message, typ, param, code, e)
+          case ServiceUnavailable                           => ServiceUnavailableError(message, typ, param, code, e)
+          case _                                            => APIError(message, typ, param, code, e)
+        }
+      case DeserializationException(_, error) => throw new json.DeserializationException(error)
+    }
+}
+
+object OpenAISyncClient {
+  def apply(authToken: String) = new OpenAISyncClient(authToken, DefaultSyncBackend())
```
adamw, line 445: here we also need to expose a `close` method (which closes the backend if it has been created when creating the client - see sttp, it has a similar mechanism)

README.md
```diff
@@ -39,17 +39,16 @@ OpenAI API Offical Documentation https://platform.openai.com/docs/api-reference/
...
+import sttp.openai.OpenAISyncClient
 import sttp.openai.requests.completions.chat.ChatRequestResponseData.ChatResponse
 import sttp.openai.requests.completions.chat.ChatRequestBody.ChatBody
 import sttp.openai.requests.completions.chat.Message
 
-// Create an instance of OpenAI providing your API secret-key
-
-val openAi: OpenAI = new OpenAI("your-secret-key")
+// Create an instance of OpenAISyncClient providing your API secret-key and sync backend instance
+val backend: SyncBackend = DefaultSyncBackend()
```
adamw, line 48: this step can be ommited, by creating the client using the default-backend factory method

README.md
```diff
@@ -62,17 +61,9 @@ val chatRequestBody: ChatBody = ChatBody(
   messages = bodyMessages
 )
 
-// Use createChatCompletion and pass created request body to create sttp request
-
-val request = openAI.createChatCompletion(chatRequestBody)
+val chatResponse = openAI.createChatCompletion(chatRequestBody)
```
adamw, line 64: and now that we have a simple type, we can add a type here :)

README.md
```diff
@@ -88,7 +79,66 @@ println(response)
 )
 */
 ```
+#### Currently only two backend implementations are available:
+* `OpenAISyncBackend` which uses `Identity` as an effect `F[_]`
+* `OpenAI` which provides raw sttp `Request`s. 
+If you want to make use of other effects, you have to use `OpenAI` and pass the chosen backend directly to `request.send(backend)` function.
```
adamw, line 85: + mention custom error handling

core/src/main/scala/sttp/openai/OpenAIErrors.scala
```diff
@@ -0,0 +1,71 @@
+package sttp.openai
+
+import sttp.client4.HttpError
+
+object OpenAIErrors {
+  sealed abstract class OpenAIError(
```
adamw, line 6: maybe let's make it simply an `OpenAIException`?

adamw: But good idea to represent open ai errors as their own model :) Maybe we should use the same approach in `OpenAI` as well?

adamw (changes requested):
Good direction, but needs some changes :)

core/src/main/scala/sttp/openai/OpenAISyncClient.scala
benzwreck: @adamw I created one in `class OpenAISyncClient`, but forgot to make docs for that (line 420):
```scala
def close(): Unit = backend.close()
```
I don't know if we need that in companion object tbh and if so how should that look like :)

adamw: The problem is we should only close the backend if we opened it (if it has been provided externally, `close` should be a no-op). See here: https://github.com/softwaremill/sttp/blob/master/core/src/main/scalajvm/sttp/client4/httpclient/HttpClientSyncBackend.scala#L21

README.md
```diff
@@ -62,33 +60,84 @@ val chatRequestBody: ChatBody = ChatBody(
...
+  model = "gpt-3.5-turbo",
+  messages = bodyMessages
+)
+
+val chatResponse = openAI
+  .createChatCompletion(chatRequestBody)
+  .send(backend)
+  .map(_.body)
+
+Thread.sleep(1000) // wait for the future to complete, just for a test case
```
adamw, line 118: let's use Await.result(future) instead

core/src/test/scala/sttp/openai/client/SyncClientSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+    (BadRequest, classOf[InvalidRequestException]),
+    (NotFound, classOf[InvalidRequestException]),
+    (UnsupportedMediaType, classOf[InvalidRequestException]),
+    (Unauthorized, classOf[AuthenticationException]),
+    (Forbidden, classOf[PermissionException]),
+    (Conflict, classOf[TryAgain]),
+    (ServiceUnavailable, classOf[ServiceUnavailableException]),
+    (Gone, classOf[APIException])
+  )
+  private val errorResponse =
```
kciesielski, line 26: **suggestion:** Please add a test that verifies if this gets deserialized to proper fields in the exception.

core/src/main/scala/sttp/openai/json/SnakePickle.scala
```diff
@@ -45,20 +47,52 @@ object SttpUpickleApiExtension extends SttpUpickleApi {
...
+      .mapWithMetadata { (string, metadata) =>
+        if (metadata.isSuccess) Right(string) else Left(httpToOpenAIError(HttpError(string, metadata.code)))
+      }
+      .showAs("either(as error, as string)")
+
+  private def httpToOpenAIError(he: HttpError[String]): OpenAIException = {
+    import sttp.model.StatusCode._
+    val errorMessageBody = SnakePickle.read[ujson.Value](he.body).apply("error").obj.values
+    val (message, typ, param, code) =
+      SnakePickle.read[(Option[String], Option[String], Option[String], Option[String])](errorMessageBody)
```
kciesielski, line 86: **issue:** You are using a tuple here, which means that you assume fields based by their position, right? If that's the case then I think it's unsafe, because the ordering is  not guaranteed and all of them are strings. It's better to deserialize to a case class.

adamw (approved):
Ship it! :)

kciesielski (approved):

PR #79, opened 2023-05-12
Update README

benzwreck:


adamw:
We need to make this checked by mdoc ;)

PR #80, opened 2023-05-12
Add mdoc

benzwreck:
Add mdoc code compilation check to docs(in our case only to README.md ;))

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,7 +26,6 @@ lazy val core = (projectMatrix in file("core"))
   )
   .settings(commonSettings: _*)
 
-//TODO this should be invoked by compilation process, see #https://github.com/scalameta/mdoc/issues/355
 val compileDocs: TaskKey[Unit] = taskKey[Unit]("Compiles docs module throwing away its output")
 compileDocs := {
   (docs.jvm(scala2.head) / mdoc).toTask(" --out target/sttp-openai-docs").value
--- a/build.sbt
+++ b/build.sbt
@@ -25,3 +25,28 @@ lazy val core = (projectMatrix in file("core"))
     libraryDependencies ++= Seq(Libraries.uPickle) ++ Libraries.sttpClient ++ Seq(Libraries.scalaTest)
   )
   .settings(commonSettings: _*)
+
+//TODO this should be invoked by compilation process, see #https://github.com/scalameta/mdoc/issues/355
+val compileDocs: TaskKey[Unit] = taskKey[Unit]("Compiles docs module throwing away its output")
+compileDocs := {
+  (docs.jvm(scala2.head) / mdoc).toTask(" --out target/sttp-openai-docs").value
+}
+
+lazy val docs = (projectMatrix in file("generated-docs")) // important: it must not be docs/
+  .enablePlugins(MdocPlugin)
+  .settings(commonSettings)
+  .settings(
+    mdocIn := file("README.md"),
+    moduleName := "sttp-openai-docs",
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -25,3 +25,28 @@ lazy val core = (projectMatrix in file("core"))
...
+  .settings(
+    mdocIn := file("README.md"),
+    moduleName := "sttp-openai-docs",
+    mdocOut := file("generated-docs/out"),
+    mdocExtraArguments := Seq("--clean-target"),
+    publishArtifact := false,
+    name := "docs",
+    libraryDependencies ++= Seq(
+      "com.softwaremill.sttp.client4" %% "cats" % "4.0.0-M1",
+      "org.typelevel" %% "cats-effect" % "3.6-1f95fd7"
```
adamw, line 47: 3.6? I think 3.5.0 is the latest stable :) but sttp-client4-cats shoudl bring in the necessary dep

build.sbt
```diff
@@ -25,3 +25,28 @@ lazy val core = (projectMatrix in file("core"))
...
   )
   .settings(commonSettings: _*)
+
+//TODO this should be invoked by compilation process, see #https://github.com/scalameta/mdoc/issues/355
+val compileDocs: TaskKey[Unit] = taskKey[Unit]("Compiles docs module throwing away its output")
+compileDocs := {
+  (docs.jvm(scala2.head) / mdoc).toTask(" --out target/sttp-openai-docs").value
+}
+
+lazy val docs = (projectMatrix in file("generated-docs")) // important: it must not be docs/
```
adamw, line 35: doesn't `generated-docs` need to be in `.gitignore`?

adamw: ah you are overwriting the out, do you need to specify both `--out` and `mdocOut` then?

adamw:
WIll be great to have this - thanks :)

build.sbt
flsh86: Is this dependecy required, for mdoc? 
```
"org.typelevel" %% "cats-effect" % "3.6-1f95fd7"
```

benzwreck: Unfortunately, it doesn't
```scala
error: README.md:93:1: object ExitCode is not a member of package cats.effect
import cats.effect.{ExitCode, IO, IOApp}
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
error: README.md:101:21: not found: type IOApp
object Main extends IOApp {
                    ^^^^^
error: README.md:102:41: not found: type IO
  override def run(args: List[String]): IO[ExitCode] = {
                                        ^^
error: README.md:116:36: not found: type IO
    HttpClientCatsBackend.resource[IO]().use { backend =>
                                   ^^
```
and more. But you are right, should have used latest stable `3.5.0` :)

adamw: Ah, the dependency in sttp is only on the kernel. So you'll need it, yeah

build.sbt
benzwreck: I guess if we want to check readme's code snippets compile execution only, we could set `mdocOut` to` --out target/sttp-openai-docs` so the output will not affect the repo and be ignored by git. Wdyt?

adamw: yes putting this into `target` sounds good :) I would also remote the TODOs as I think they don't apply here

kciesielski (approved):

build.sbt
adamw: bump :) this seems to be copied from tapir, along with comments which don't necessarily make sense here

Issue #59, opened 2023-04-27
Create chat completions SSE


benzwreck:
Create step SSE endpoint
https://platform.openai.com/docs/api-reference/chat/create

PR #124, opened 2023-11-06
Add streaming support for chat completions

DybekK:
Closes https://github.com/softwaremill/sttp-openai/issues/59

This PR adds streaming support for [Create chat completion](https://platform.openai.com/docs/api-reference/chat/create) endpoint using server-sent events.

_**Note**_: For the first version we used `Fs2` streams, we'll support other stream implementations in the feature.

Here are the main changes and enhancements made in this PR:

- case classes related to chat completions have been redefined to align with the latest OpenAI API. This ensures compatibility and consistency with the OpenAI API.
- `OpenAI` class has been expanded with new `createChatCompletion` method. This method is designed to work with generic stream implementations using `Streams[S]`
- **out-of-the-box** streaming support has been added using the `Fs2` library by extending current `OpenAI` class using implicit class:
```scala
package sttp.openai.streaming.fs2

object Fs2OpenAI {
  implicit class streaming(val client: OpenAI) {
    def createStreamedChatCompletion[F[_]: RaiseThrowable](chatBody: ChatBody): StreamRequest[Either[OpenAIException, Stream[F, ChatChunkResponse]], Fs2Streams[F]]
  }
}
```
- README has been updated to include an example demonstrating how to use this new feature
- integration tests have been added using `HttpClientFs2Backend.stub`


streaming/fs2/src/main/scala/sttp/openai/streaming/fs2/Fs2OpenAI.scala
```diff
@@ -0,0 +1,48 @@
...
+import sttp.client4.StreamRequest
+import sttp.client4.impl.fs2.Fs2ServerSentEvents
+import sttp.model.sse.ServerSentEvent
+import sttp.openai.OpenAI
+import sttp.openai.OpenAIExceptions.OpenAIException
+import sttp.openai.json.SttpUpickleApiExtension.deserializeJsonSnake
+import sttp.openai.requests.completions.chat.ChatChunkRequestResponseData.ChatChunkResponse
+import sttp.openai.requests.completions.chat.ChatRequestBody.ChatBody
+
+object Fs2OpenAI {
```
kciesielski, line 14: ** suggestion ** How about creating an object called `fs2` in package `sttp.openai.streaming`. This would make the official required import be just a nice `import sttp.openai.streaming.fs2._`.

core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala
```diff
@@ -45,7 +63,8 @@ object ChatRequestBody {
       presencePenalty: Option[Double] = None,
       frequencyPenalty: Option[Double] = None,
       logitBias: Option[Map[String, Float]] = None,
-      user: Option[String] = None
+      user: Option[String] = None,
+      stream: Option[Boolean] = None
```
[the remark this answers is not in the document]
DybekK: You're right, it might be confusing. I completely removed the stream parameter from `ChatBody` and created `ChatBody.withStreaming` method which directly adds `"stream": true` field to json when the createChatCompletion with streams is invoked.

streaming/fs2/src/main/scala/sttp/openai/streaming/fs2/Fs2OpenAI.scala
DybekK: It's problematic, an object named `fs2` breaks imports from actual fs2 library.

kciesielski: Imports from `fs2` are only clashing within sttp-openai itself, not for users. You can resolve this by using

```scala
import _root_.fs2.whatever
```
and this is a common pattern.

PR #127, opened 2023-11-07
Add ZIO streaming support for chat completions

DybekK:
This PR adds streaming support for [Create chat completion](https://platform.openai.com/docs/api-reference/chat/create) endpoint using ZIO as a streams implementation.

Here are the main changes made in this PR:

- **out-of-the-box** streaming support has been added using the ZIO library by extending current OpenAI class using implicit class:
```scala
package sttp.openai.streaming

package object zio {
  implicit class extension(val client: OpenAI) {
    def createStreamedChatCompletion(chatBody: ChatBody): StreamRequest[Either[OpenAIException, Stream[Throwable, ChatChunkResponse]], ZioStreams]
  }
}
```
- integration tests have been added using HttpClientZioBackend.stub

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ To use `Fs2` add the following import:
 import sttp.openai.streaming.fs2._
 ```
 
-To use `Zio` add the following import:
+To use `ZIO` add the following import:
 
 ```scala
 import sttp.openai.streaming.zio._
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ object Main extends App {
   */
 }
 ```
-#### Currently only two backend implementations are available:
+#### Available backend implementations:
 * `OpenAISyncBackend` which uses identity monad `Id[A]` as an effect `F[A]` and throws `OpenAIException`
 * `OpenAI` which provides raw sttp `Request`s and wraps `Response`s into `Either[OpenAIException, A]`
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -140,14 +140,20 @@ object Main extends IOApp {
 ```
 
 #### Create completion with streaming:
-The Chat Completion API features streaming support via server-sent events. Currently, we only support streaming using `Fs2`. 
+The Chat Completion API features streaming support via server-sent events. Currently, we only support streaming using `Fs2` and `Zio`
```
kciesielski, line 143: ** nitpick ** `Zio` -> `ZIO`

build.sbt
```diff
@@ -40,6 +41,16 @@ lazy val fs2 = (projectMatrix in file("streaming/fs2"))
...
 
+lazy val zio = (projectMatrix in file("streaming/zio"))
+  .jvmPlatform(
+    scalaVersions = scala2 ++ scala3
+  )
+  .settings(commonSettings)
+  .settings(
+    libraryDependencies += Libraries.sttpClientZio
+  )
+  .dependsOn(core % "compile->compile;test->test")
```
kciesielski, line 52: ** question ** What do we need `test->test` for? I suppose it's needed for
```scala
import sttp.client4.testing.RawStream
```
?

DybekK: It allows you to depend on test dependencies from core module. In our case we use this to access ScalaTest library.

kciesielski: `test->test` also means that `src/test` stuff from `core` can be used in `tests` of this module. In this case I checked and it's indeed needed for common test stuff like `sttp.openai.fixtures.*`

kciesielski (approved):

build.sbt
DybekK: Right, I forgot to mention it 😅

PR #128, opened 2023-11-07
Add Pekko streaming support for chat completions

DybekK:
This PR adds streaming support for [Create chat completion](https://platform.openai.com/docs/api-reference/chat/create) endpoint using Pekko Streams as a streams implementation.

Here are the main changes made in this PR:
- **out-of-the-box** streaming support has been added using the Pekko Streams library by extending current OpenAI class.
- integration tests have been added using `PekkoHttpBackend.stub`

_Minor change_:
- `DeserializationException` was replaced with more detailed `DeserializationOpenAIException` during tests

### DIFF
--- a/streaming/pekko/src/main/scala/sttp/openai/streaming/pekko/package.scala
+++ b/streaming/pekko/src/main/scala/sttp/openai/streaming/pekko/package.scala
@@ -1,0 +1,51 @@
+package sttp.openai.streaming
+
+import org.apache.pekko.stream.scaladsl.{Flow, Source}
+import org.apache.pekko.util.ByteString
+import sttp.capabilities.pekko.PekkoStreams
+import sttp.client4.StreamRequest
+import sttp.model.sse.ServerSentEvent
+import sttp.openai.OpenAI
+import sttp.openai.OpenAIExceptions.OpenAIException
+import sttp.openai.json.SttpUpickleApiExtension.deserializeJsonSnake
+import sttp.openai.requests.completions.chat.ChatChunkRequestResponseData.ChatChunkResponse
+import sttp.openai.requests.completions.chat.ChatRequestBody.ChatBody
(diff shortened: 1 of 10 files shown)


streaming/pekko/src/main/scala/sttp/openai/streaming/pekko/package.scala
```diff
@@ -0,0 +1,51 @@
...
+      response: Either[OpenAIException, Source[ByteString, Any]]
+  ): Either[OpenAIException, Source[ChatChunkResponse, Any]] =
+    response.map(
+      _.via(PekkoHttpServerSentEvents.parse)
+        .via(deserializeEvent)
+    )
+
+  private def deserializeEvent: Flow[ServerSentEvent, ChatChunkResponse, Any] =
+    Flow[ServerSentEvent].collect {
+      case ServerSentEvent(Some(data), _, _, _) if data != DoneEventMessage =>
```
kciesielski, line 45: ** issue ** This makes me thinking - the `DoneEventMessage` is simply filtered here, and we basically rely on the stream end to be signalised. Maybe we should rather `takeWhile` instead of `collect`, which would guarantee that `DoneEventMessage` completes the stream right away.
For that, we may need a test where there are more messages after `DoneEventMessage`, it's unexpected but we would be sure that "done" means "done" for us, no matter what happens.

kciesielski: Similarly with `fs2` and `zio`, these modules may also need such an improvement.

DybekK: Thanks for that suggestion, I added takeWhile to the Pekko implementation

kciesielski (approved):

PR #136, opened 2023-11-17
Redefine chat completion case classes

DybekK:
This PR redefines case classes for [Create chat completion](https://platform.openai.com/docs/api-reference/chat/create) to align with the latest OpenAI API.


build.sbt
```diff
@@ -88,10 +88,6 @@ lazy val docs = (projectMatrix in file("generated-docs")) // important: it must
     mdocExtraArguments := Seq("--clean-target"),
     publishArtifact := false,
     name := "docs",
-    libraryDependencies ++= Seq(
-      "com.softwaremill.sttp.client4" %% "cats" % "4.0.0-M6",
-      "org.typelevel" %% "cats-effect" % "3.5.2"
-    ),
```
amorfis, line 94: We keep all the dependencies in `Dependencies.scala` and I'd rather move it there too

DybekK: Currently, all dependencies are in `Dependencies.scala`. I've removed `libraryDependencies` from the docs module as they are now derived from other modules, in this case, from `fs2` module.

core/src/main/scala/sttp/openai/requests/completions/chat/ToolCall.scala
```diff
@@ -0,0 +1,16 @@
...
+import sttp.openai.json.SnakePickle
+
+/** @param id
+  *   The ID of the tool call.
+  * @param type
+  *   The type of the tool. Currently, only function is supported.
+  * @param function
+  *   The function that the model called.
+  */
+case class ToolCall(id: String, `type`: String, function: FunctionCall)
```
amorfis, line 12: If in future we plan to implement also tools other than "function" - maybe `ToolCall` should be a trait, and we should have `FunctionToolCall` instead of this case class, implementing `ToolCall` trait?

build.sbt
amorfis: Was it code removed, when I thought it's code added?... Ignore me 🤦

core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala
```diff
@@ -3,63 +3,74 @@ package sttp.openai.requests.completions.chat
...
-    *   Number between -2.0 and 2.0. Positive values penalize new tokens based on their existing frequency in the text so far, decreasing
-    *   the model's likelihood to repeat the same line verbatim.
-    * @param logitBias
-    *   Modify the likelihood of specified tokens appearing in the completion.
+    * @param tools
+    *   A list of tools the model may call. Currently, only functions are supported as a tool. Use this to provide a list of functions the
+    *   model may generate JSON inputs for.
+    * @param toolChoice
+    *   Controls which (if any) function is called by the model. "none" means the model will not call a function and instead generates a
+    *   message. "auto" means the model can pick between generating a message or calling a function. Specifying a particular function via
```
kciesielski, line 53: The first sentence is OK, but the individual subtypes should be described in the documentation of `ToolChoice` type. Additionaly, You are describing JSON structure, while it would be better to describe possibilities according to our types. 
Can we use a sealed trait instead of `AsString`? Something like
```scala
sealed trait ToolChoice
case class ToolNone() extends ToolChoice
case class ToolAuto() extends ToolChoice
case class ToolFunction(name: String) extends ToolChoice
```
and then a custom (de)serializer which uses `type` field as discriminator?

core/src/main/scala/sttp/openai/requests/completions/chat/message/Message.scala
```diff
@@ -0,0 +1,83 @@
...
+import sttp.openai.json.SnakePickle
+import sttp.openai.requests.completions.chat.ToolCall
+import ujson._
+
+sealed trait Message
+
+object Message {
+  case class SystemMessage(content: String, name: Option[String] = None) extends Message
+  case class UserMessage(content: Content, name: Option[String] = None) extends Message
+  case class AssistantMessage(content: String, name: Option[String] = None, toolCalls: Option[Seq[ToolCall]] = None) extends Message
```
kciesielski, line 12: ** nitpick ** `Option[Seq[T]]` is sometimes confusing for users, as it is difficult to guess the difference between `None` and `Some(Nil)`. If there's no difference in this case, then maybe we can make it even more convenient for the user and put just a `Seq[T] = Nil` here, which is treated as `None` by the serializer if empty?

core/src/main/scala/sttp/openai/requests/completions/chat/message/Tool.scala
```diff
@@ -0,0 +1,24 @@
+package sttp.openai.requests.completions.chat.message
+
+import sttp.openai.json.SnakePickle
+import ujson.Value
+
+/** @param type
+  *   The type of the tool. Currently, only function is supported.
+  */
+case class Tool(`type`: String, function: Tool.FunctionCall)
```
kciesielski, line 9: How about creating a `sealed trait Tool` with one case class `FunctionTool`, then use `type` as a discriminator field?

core/src/test/scala/sttp/openai/utils/ChatCompletionUtils.scala
```diff
@@ -0,0 +1,72 @@
+package sttp.openai.utils
+
+import sttp.openai.requests.completions.chat.{FunctionCall, ToolCall}
+import sttp.openai.requests.completions.chat.message._
+import ujson._
+
+object ChatCompletionUtils {
```
kciesielski, line 7: ** nitpick ** I'd rename this to `ChatCompletionFixtures`

core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala
```diff
@@ -3,63 +3,74 @@ package sttp.openai.requests.completions.chat
...
+    *   `{"type": "function", "function": {"name": "my_function"}}` forces the model to call that function. "none" is the default when no
+    *   functions are present. "auto" is the default if functions are present.
     * @param user
     *   A unique identifier representing your end-user, which can help OpenAI to monitor and detect abuse.
     */
   case class ChatBody(
```
kciesielski, line 59: ** Question ** I'm not sure if sorting arguments alphabetically is a good idea. It may be difficult for current users to find out how they should update their requests with new fields. I would just add new fields to the end of argument list, so that IDE may be more helpful with finding out what should be changed. 
In the future, when new arguments are added, they should also go to the end of the class. Alphabetical order isn't very helpful anyway, right?
Or maybe it's the way OpenAI defines their arguments in the API and we want to have matching models for easier comparison and documentation reading overall?

DybekK: > Or maybe it's the way OpenAI defines their arguments in the API and we want to have matching models for easier comparison and documentation reading overall?

Exactly this one

kciesielski: Ok, in that case sorted fields may be justified :)

core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala
```diff
@@ -3,63 +3,74 @@ package sttp.openai.requests.completions.chat
...
-    * @param frequencyPenalty
-    *   Number between -2.0 and 2.0. Positive values penalize new tokens based on their existing frequency in the text so far, decreasing
-    *   the model's likelihood to repeat the same line verbatim.
-    * @param logitBias
-    *   Modify the likelihood of specified tokens appearing in the completion.
+    * @param tools
+    *   A list of tools the model may call. Currently, only functions are supported as a tool. Use this to provide a list of functions the
+    *   model may generate JSON inputs for.
+    * @param toolChoice
+    *   Controls which (if any) function is called by the model. "none" means the model will not call a function and instead generates a
```
kciesielski, line 52: ** suggestion ** The new `ToolChoice` sealed hierarchy looks really good. I think you can leave just the first sentence, without explaining individual subtypes. Then, in actual subtypes of `ToolChoice`, add a scaladoc for each case, like
```scala
/** Means that the model will not call a function and instead generates a message. */
case object ToolNone extends ToolChoice
// etc
```
(Yes, my next suggestion is to make `ToolNone` a case object, but that's in another comment).

core/src/main/scala/sttp/openai/requests/completions/chat/ToolCall.scala
kciesielski: +1, and use the `type` field as a discriminator.

kciesielski (approved):

guersam:
In Scala 3, missing `toolCalls` causes a deserialization error when there are no tools:

```scala
OpenAISyncClient(secret).createChatCompletion(ChatBody(
  model = ChatCompletionModel.GPT4Turbo,
  messages = Seq(
    Message.UserMessage(
      content = Content.TextContent("Hello!")
    )
  )
))
```

```
Fail(sttp.openai.OpenAIExceptions$OpenAIException$DeserializationOpenAIException: missing keys in dictionary: toolCalls at index 352
```

guersam:
Opened #137.

PR #137, opened 2023-11-24
Fix chat reponse deserialization with no toolCalls

guersam:
Currently, missing `toolCalls` causes a deserialization error when there are no tools:

```scala
OpenAISyncClient(secret).createChatCompletion(ChatBody(
  model = ChatCompletionModel.GPT4Turbo,
  messages = Seq(
    Message.UserMessage(
      content = Content.TextContent("Hello!")
    )
  )
))
```

```
Fail(sttp.openai.OpenAIExceptions$OpenAIException$DeserializationOpenAIException: missing keys in dictionary: toolCalls at index 352
```

This commit sets default value for `toolCalls` to avoid the error.


DybekK:
Thank you for your contribution 😃

guersam:
My bad. At least in Scala 3, even with the default function argument, it still fails to deserialize when there are no `toolCalls` field in the JSON response :cry:

PR #138, opened 2023-11-24
Remove null from streaming finishReason

guersam:
Currently, initial response chunks from chat completion emits `null` at `choices[*].finishReason`:

```scala
ChatChunkResponse(chatcmpl-8OGfnT5mmJwiTTmICOgi26fOxrJnf,List(Choices(Delta(Some(Assistant),Some(),None,None),null,0)),1700794655,gpt-4-1106-preview,chat.completion.chunk,Some(fp_a24b4d720c))
<FAIL>: Die(java.lang.NullPointerException: Cannot invoke "String.length()" because the return value of "sttp.openai.requests.completions.chat.ChatChunkRequestResponseData$Choices.finishReason()" is null,Stack trace for thread "zio-fiber-":
)
```

This PR makes it optional.


guersam:
Just noticed that `toolCalls.id` should be optional as well in streaming mode.

Issue #151, opened 2024-02-20
Support for OpenAI BETA API - Assistants, Threads, etc.


dgolda:
I'm adding support for Beta features like Assistants and Threads.

https://platform.openai.com/docs/api-reference/assistants

I'm wondering if the new methods should only be added to OpenAI and OpenAISyncClient, or should they be in a separate trait or class?

Any tips?

My work in progress:
https://github.com/dgolda/sttp-openai/tree/151-support-for-assistants

I noticed that the class naming is inconsistent - which names can be considered exemplary?

adamw:
Ah that's great, the OpenAI API is in constant flux :)

Yes, I think these are the two places that need modifying. Plus any required payload classes in the `requests` package

DybekK:
Thank you for your contribution! Your pull request has been merged :)

PR #152, opened 2024-02-23
#151 Support for Assistants, Threads, Messages and Runs

dgolda:
Added support for BETA OpenAI features: Assistants, Threads (with Messages), Runs (with RunSteps).

https://platform.openai.com/docs/api-reference/assistants


DybekK (approved):
LGTM!

PR #169, opened 2024-05-02
Added non-openAI backends support

### LINKED ISSUE #166: Add support for OpenAI API-compatible servers
Given that OpenAI API shape became a de-facto standard in the industry it would be beneficial to add support for pluggable base URLs. Currently base urls are hardcoded into the library.

Here's a quick list of projects that are, to some degree of course, api-compatible with OpenAI:

* https://console.groq.com/docs/openai
* https://docs.together.ai/docs/openai-api-compatibility
* https://github.com/lm-sys/FastChat/blob/main/docs/openai_api.md
* https://github.com/ggerganov/llama.cpp/tree/master/examples/server
* https://ollama.com/blog/openai-compatibility 
* https://github.com/nomic-ai/gpt4all/blob/main/gpt4all-api/README.md
* https://lmstudio.ai/docs/local-server

Most of these are quite significant with Groq being the fastest inference backend on the market, ollama and llama.cpp being de-facto standard engines for local CPU/GPU inference.

Might find some time to work on this.

lbialy:
Closes #166 

BaseUri is *intentionally* made for user to include **`/v1/`**, check out the use of `openai` library in python:

```python
from openai import OpenAI

client = OpenAI(
    base_url = 'http://localhost:11434/v1',
    api_key='ollama', # required, but unused
)
```

compare:

```scala
val client = OpenAISyncClient("ollama", uri"http://localhost:11434/v1")
```

**Things left to figure out:**
* Lots of APIs do not work, most of the time what does work is just `/v1/chat/completions`, `/v1/completions` and if weather, moon phase, time of day and horoscopes are right - `/v1/embeddings` too. Ollama doesn't implement `/v1/models` but Grok does and so on.
* Error modes are slightly different. When other companies / foss orgs say that API is openAI-compatible it doesn't mean it handles errors the same way. For example: ollama behaves fine and if you use a non-existent model you get a nice:
```
sttp.openai.OpenAIExceptions$OpenAIException$InvalidRequestException: statusCode: 404, response: {"error":{"message":"model 'mixtral-8x7b-32768' not found, try pulling it first","type":"api_error","param":null,"code":null}}
```
which is sort of what I'd expect (and what `sttp-openai` would expect too). For Grok however that's not as nice because their error json doesn't contain one of the expected fields, apparently:
```scala
scala> println(oai.createChatCompletion(ChatBody(model = ChatCompletionModel.CustomChatCompletionModel("mistral"), messages = Seq(Message.UserMessage(content = Content.TextContent("What is Scala programming language?"))))).choices.head.message.content.trim)
upickle.core.Abort: missing keys in dictionary: param
```

```bash
λ curl https://api.groq.com/openai/v1/chat/completions -w "%{http_code}" \
        -H 'Content-Type: application/json' \
        -d '{
          "model": "mistral",
          "messages": [
            {
              "role": "user",
              "content": "How do I shot web?"
            }
          ]
    }'
{"error":{"message":"Invalid API Key","type":"invalid_request_error","code":"invalid_api_key"}}
401
```
```bash
λ curl https://api.groq.com/openai/v1/chat/completions -w "%{http_code}" \
        -H 'Content-Type: application/json' \
        -H 'Authorization: Bearer REDACTED' \
        -d '{
          "model": "mistral",
          "messages": [
            {
              "role": "user",
              "content": "How do I shot web?"
            }
          ]
    }'
{"error":{"message":"The model `mistral` does not exist or you do not have access to it.","type":"invalid_request_error","code":"model_not_found"}}
404
```

Now this is probably another issue with upickle because `param` field is usually defined as `param: Option[String]` and when it's not present in the json is should be deserialized to None. Maybe we should configure upickle to be more permissive in error json parsing?

### DIFF
--- a/core/src/main/scala/sttp/openai/OpenAI.scala
+++ b/core/src/main/scala/sttp/openai/OpenAI.scala
@@ -1084,27 +1084,27 @@ class OpenAI(authToken: String, baseUri: Uri = OpenAIUris.OpenAIBaseUri) {
 
 }
 
-private class OpenAIUris(val BaseUri: Uri) {
-  private val ImageBase: Uri = uri"$BaseUri/images"
-  private val AudioBase: Uri = uri"$BaseUri/audio/"
+private class OpenAIUris(val baseUri: Uri) {
+  private val imageBase: Uri = uri"$baseUri/images"
+  private val audioBase: Uri = uri"$baseUri/audio/"
 
-  val ChatCompletions: Uri = uri"$BaseUri/chat/completions"
-  val Completions: Uri = uri"$BaseUri/completions"
-  val CreateImage: Uri = ImageBase.addPath("generations")
-  val Edits: Uri = uri"$BaseUri/edits"
-  val Embeddings: Uri = uri"$BaseUri/embeddings"
-  val EditImage: Uri = ImageBase.addPath("edits")
-  val Files: Uri = uri"$BaseUri/files"
-  val FineTunes: Uri = uri"$BaseUri/fine-tunes"
-  val Models: Uri = uri"$BaseUri/models"
-  val Moderations: Uri = uri"$BaseUri/moderations"
-  val Transcriptions: Uri = AudioBase.addPath("transcriptions")
-  val Translations: Uri = AudioBase.addPath("translations")
-  val VariationsImage: Uri = ImageBase.addPath("variations")
+  val ChatCompletions: Uri = uri"$baseUri/chat/completions"
+  val Completions: Uri = uri"$baseUri/completions"
+  val CreateImage: Uri = imageBase.addPath("generations")
+  val Edits: Uri = uri"$baseUri/edits"
+  val Embeddings: Uri = uri"$baseUri/embeddings"
+  val EditImage: Uri = imageBase.addPath("edits")
+  val Files: Uri = uri"$baseUri/files"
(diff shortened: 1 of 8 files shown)


lbialy:
I've fixed missing field error serialization issue in last commit. Hopefully we won't see many other error channel issues with this.

core/src/main/scala/sttp/openai/OpenAI.scala
```diff
@@ -1082,27 +1084,27 @@ class OpenAI(authToken: String) {
 
 }
 
-private object OpenAIUris {
-  private val ImageBase: Uri = uri"https://api.openai.com/v1/images"
-  private val AudioBase: Uri = uri"https://api.openai.com/v1/audio/"
+private class OpenAIUris(val BaseUri: Uri) {
```
adamw, line 1087: [minor] shouldn't this be lower-case? `baseUri`

lbialy: could be, in this case also `imageBase` and `audioBase`, right?

adamw: `ImageBase` in the old code was a constant, this is a variable, so it's written using lowercase I guess? :)

adamw:
Thanks :)

PR #189, opened 2024-07-26
Support gpt-4o-mini

nemoo:
Support gpt-4o-mini
Use gpt-4o-mini as default in README
Add info about CustomChatCompletionModel in ChatGPT section


adamw:
Thanks!

PR #217, opened 2024-10-10
Implement JSON schema support with apispec

### LINKED ISSUE #216: JSON schema support
Would you consider adding support for the [`json_schema` response format](https://platform.openai.com/docs/guides/structured-outputs/how-to-use?context=without_parse)? If so, I would be happy to take a stab at a pull request to add it. I would model it similarly to this: https://github.com/cequence-io/openai-scala-client/blob/master/openai-core/src/main/scala/io/cequence/openaiscala/domain/JsonSchema.scala

mrdziuban:
Fixes #216 

@adamw re: your suggestion:

> An extension to this idea would be to derive a Tapir schema from a case class, and then [serialize it as JSON Schema](https://tapir.softwaremill.com/en/latest/docs/json-schema.html).

I chose to use the `Schema` type from [sttp-apispec](https://github.com/softwaremill/sttp-apispec) to implement this, but I didn't add explicit support for passing a Tapir schema and having it be converted automatically -- I figured downstream users could do the conversion themselves if they wanted to.

OpenAI only [supports a subset of JSON schema](https://platform.openai.com/docs/guides/structured-outputs/supported-schemas) and has some limitations, which I described in a comment in `ChatRequestBody.scala`.

I don't think it's reasonable to enforce all of these constraints in this library, users will just have to be cognizant of whether their schema violates any of them. If it does, the API returns an error like this:

```json
{
  "error": {
    "message": "Invalid schema for response_format 'schemaName': ... reason why schema was invalid ...",
    "type": "invalid_request_error",
    "param": "response_format",
    "code": null
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,6 +427,119 @@ object Main extends OxApp:
 
 See also the [ChatProxy](https://github.com/softwaremill/sttp-openai/blob/master/examples/src/main/scala/examples/ChatProxy.scala) example application.
 
+#### Structured Outputs/JSON Schema support
+
+To take advantage of [OpenAI's Structured Outputs](https://platform.openai.com/docs/guides/structured-outputs/introduction)
+and support for JSON Schema, you can use `ResponseFormat.JsonSchema` when creating a completion.
+
+The example below produces a JSON object:
+
+```scala mdoc:compile-only
+//> using dep com.softwaremill.sttp.openai::core:0.2.3
+
+import scala.collection.immutable.ListMap
+import sttp.apispec.{Schema, SchemaType}
+import sttp.openai.OpenAISyncClient
+import sttp.openai.requests.completions.chat.ChatRequestResponseData.ChatResponse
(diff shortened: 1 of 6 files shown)


mrdziuban:
Ah I hadn't seen #205, this definitely overlaps with that. I think implementing the support with apispec is more robust than just a `ujson.Value` as in #205, and it leaves less room for error. My changes don't add `strict` to `FunctionTool` like #205 does though

adamw:
This looks great! Could you add an example to the readme (also maybe mentioning how to auto-derive the schema using tapir)? That would help a lot when it comes to discoverability of the feature. Alternatively, there's the `examples` sub-project, to which you simply link from the readme

mrdziuban:
Definitely, updated the readme to show both manual and derived construction of JSON schemas.

I also discovered and fixed a bug while testing schema derivation where existing `required` and `additionalProperties` fields on the schema took precedence over the ones we add ourselves, which resulted in invalid schema errors from the API.

adamw:
Awesome - thanks! :)

PR #241, opened 2024-12-06
Allow for discriminated unions in OpenAPI schema

pstutz:
This PR enhances the schema transformation logic by recognizing discriminated unions and skipping the forced `additionalProperties: false` constraint for them. As a result, complex, variant-based schemas remain flexible and validate correctly. Testing against the OpenAI API confirms that this approach is supported.

### DIFF
--- a/core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala
+++ b/core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala
@@ -31,18 +31,17 @@ object ChatRequestBody {
             requiredProperties: List[String]
         )
 
-        /** OpenAI's JSON schema support has a couple limitations/requirements we need to account for:
+        /** OpenAI's JSON schema support imposes two requirements:
           *
-          *   1. All fields must be `required` https://platform.openai.com/docs/guides/structured-outputs/all-fields-must-be-required
+          *   1. All fields must be `required`: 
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! Is there a chance for a test maybe?

pstutz:
The approach I'm suggesting removes an existing, untested workaround that forces `additionalProperties=false` when it is unnecessary. I've verified through testing that OpenAI accepts discriminated unions without this property, but there's no spec for this behavior.

The current implementation in ChatRequestBody [automatically applies provider-specific schema transformations](https://github.com/softwaremill/sttp-openai/blob/ca4b48f636ca49e6932f2cd3e6bf03fbfc130579/core/src/main/scala/sttp/openai/requests/completions/chat/ChatRequestBody.scala#L47). Given that different API providers may have varying expectations, a more robust approach might be to allow users to supply their own *final* OpenAPI schema for structured outputs, rather than automatically applying provider-specific workarounds on top of it.

I think merging the PR short-term to address a current functionality limitation and then moving away from making assumptions about provider behavior or at least offering override options would help get us to a more reliable solution faster than adding tests for undocumented provider behavior.

adamw:
Ok, thanks, let's see how it works :) Would you be able to work on the target solution? Either way, if you could create an issue describing the work to be done, that would be great.

pstutz:
Thanks for the merge. I'll create an issue for the provider specific adaptation question.

PR #242, opened 2024-12-07
Deserialize to custom model where possible

kapunga:
When using the library with a locally running instance of [Ollama](https://ollama.com/), a `DeserializationOpenAIException` is thrown when passing a custom model.

This PR deserializes to a custom model where applicable to avoid this exception.


adamw:
Thanks!

PR #243, opened 2024-12-07
Fix `Choices` Deserialization

kapunga:
Some OpenAPI Compatible APIs like Ollama don't always return `lobprobs` as part of the `choices` JSON. To handle this we need to add a default argument to `logprobs` so upickle can handle this case. A private case class is used to avoid having to change the order of arguments which would be a breaking change.


adamw:
Breaking changes are fine in 0.x versions. Also, maybe there's a some chance to write a test?

kapunga:
> Breaking changes are fine in 0.x versions. Also, maybe there's a some chance to write a test?

Sounds good. I updated and added an `ollama` deserialization test that should also cover [PR 242](https://github.com/softwaremill/sttp-openai/pull/242) as well.

grzegorz-bielski:
This would be useful to have.

Also, the `finishReason` in the `ChatChunkRequestResponseData.Choices` has the same problem.

adamw:
Sorry for being so late, somehow this fell through the cracks :)

PR #281, opened 2025-02-21
remove deprecated stuff, no longer available in OpenAI API

zygiert1990:


core/src/main/scala/sttp/openai/OpenAI.scala
```diff
@@ -567,42 +541,6 @@ class OpenAI(authToken: String, baseUri: Uri = OpenAIUris.OpenAIBaseUri) {
...
-    * [[https://platform.openai.com/docs/api-reference/fine-tunes/list]]
-    */
-  def getFineTunes: Request[Either[OpenAIException, GetFineTunesResponse]] =
-    openAIAuthRequest
-      .get(openAIUris.FineTunes)
-      .response(asJson_parseErrors[GetFineTunesResponse])
-
-  /** Immediately cancel a fine-tune job.
-    *
-    * [[https://platform.openai.com/docs/api-reference/fine-tunes/cancel]]
```
adamw, line 596: it's the first one I checked, but it seems to be there? maybe the path is different though?

https://platform.openai.com/docs/api-reference/fine-tuning/cancel

zygiert1990: Path has changed from fine-tunes to fine-tuning/jobs. Also request data has changed and most probably response. Currently I am working on add again new fine-tuning API.

zygiert1990: I meant request and response for creating of fine-tuning.

PR #291, opened 2025-03-05
add create upload method

zygiert1990:


core/src/main/scala/sttp/openai/requests/upload/UploadResponse.scala
```diff
@@ -0,0 +1,75 @@
...
+    createdAt: Int,
+    filename: String,
+    purpose: String,
+    status: String,
+    expiresAt: Int,
+    file: Option[File]
+)
+
+object UploadResponse {
+  implicit val uploadResponseR: SnakePickle.Reader[UploadResponse] = SnakePickle.macroR[UploadResponse]
```
adamw, line 39: how can you deserialise a `File` from JSON? what's the intended content of this field?

zygiert1990: Yes, this is just some file metadata :) Perhaps I should rename this class (FileMetadata) to make it clear that this is not an actual file?

adamw: ah yes, this is a good idea, let's do that :)

PR #292, opened 2025-03-07
extend Audio & Models API's

zygiert1990:


streaming/ox/src/main/scala/sttp/openai/streaming/ox/oxStreaming.scala
```diff
@@ -13,8 +13,22 @@ import sttp.openai.requests.completions.chat.ChatRequestBody.ChatBody
...
+    *
+    * @param requestBody
+    *   Request body that will be used to create a speech.
+    *
+    * @return
+    *   The audio file content.
+    */
+  def createSpeech(requestBody: SpeechRequestBody): Request[Either[OpenAIException, InputStream]] =
+    client.createSpeechAsInputStream(requestBody)
+
```
zygiert1990, line 31: Not sure does it make any sense to put here this method, as this is just delegation to client method, which doesn't do any further action.

adamw: I think it doesn't as this extension is about streaming, and there's no streaming here

adamw: (well, no ox-streaming, just regular Java InputStream)

PR #293, opened 2025-03-11
Handle nullable finish_reason in the streamed chunk completion choice

grzegorz-bielski:
`finishReason` can be null, some OpenAI protocol implementation like LiteLLM don't always return it. 

Similar to https://github.com/softwaremill/sttp-openai/pull/243


grzegorz-bielski:
Hello @adamw, please take a look.

adamw:
Thanks :)

PR #305, opened 2025-03-21
Issue 268 - Support for tool calls

zygiert1990:


README.md
```diff
@@ -539,6 +539,136 @@ val jsonSchema: ASchema = TapirSchemaToJsonSchema(
 )
 ```
 
+#### Generating JSON Schema from case class
+
+We can also generate JSON Schema directly from case class, without defining the schema manually:
```
adamw, line 544: I would expand this a bit:
* some more detail on how this differs from the example immediately above
* first mention the methods that the section focuses on (as I understand, this is about `SchematizedFunctionTool`?), describe how this works, and only later proceed to the example. The example is long and it might be hard understanding what's the main point

README.md
```diff
@@ -539,6 +539,136 @@ val jsonSchema: ASchema = TapirSchemaToJsonSchema(
...
+  implicit val bookedFlightW: SnakePickle.Writer[BookedFlight] = SnakePickle.macroW[BookedFlight]
+}
+
+object Main extends App {
+  val apiKey = System.getenv("OPENAI_KEY")
+  val openAI = OpenAISyncClient(apiKey)
+
+  val initialRequestMessage = Seq(UserMessage(content = TextContent("I want to book a flight from London to Tokyo for Jane Doe, age 34")))
+
+  val givenRequest = ChatBody(
```
adamw, line 583: let's add some comments here describing how this differs from a "normal" open ai request, and what amount of labor we are saving

README.md
```diff
@@ -539,6 +539,136 @@ val jsonSchema: ASchema = TapirSchemaToJsonSchema(
...
+          None
+        ),
+        tool_calls,
+        0,
+        None
+      )
+    )
+    */
+
+  val toolCalls = initialRequestResult.choices.head.message.toolCalls
```
adamw, line 629: here again, some context on what you're trying to achieve with that code would be good. Also earlier: what functionality are you implementing in the example? What are the actors? There's a user which specifies the prompt,  there's ChatGPT, and there's a booking service - it would be good to introduce them first

core/src/main/scala/sttp/openai/requests/completions/chat/message/Tool.scala
```diff
@@ -16,6 +20,31 @@ object Tool {
     */
   case class FunctionTool(description: String, name: String, parameters: Map[String, Value]) extends Tool
 
+  case class SchematizedFunctionTool(description: String, name: String, parameters: Schema) extends Tool
```
adamw, line 23: some scalados here would be nice, specifying when to use `FuntionTool`, and when this class, and in which API

README.md
```diff
@@ -539,6 +539,153 @@ val jsonSchema: ASchema = TapirSchemaToJsonSchema(
...
+  implicit val bookedFlightW: SnakePickle.Writer[BookedFlight] = SnakePickle.macroW[BookedFlight]
+}
+
+object Main extends App {
+  val apiKey = System.getenv("OPENAI_KEY")
+  val openAI = OpenAISyncClient(apiKey)
+
+  val initialRequestMessage = Seq(UserMessage(content = TextContent("I want to book a flight from London to Tokyo for Jane Doe, age 34")))
+
+  // Request created using SchematizedFunctionTool, all we need to do here is just define type without explicit Json Schema.
```
adamw, line 596:
```suggestion
  // Request created using SchematizedFunctionTool, all we need to do here is just define the type. The schema is automatically generated using a macro, available via the `sttp.tapir.generic.auto._` import.
```

adamw (approved):
Looks good, let's merge & release :)

Issue #321, opened 2025-04-14
listRunSteps - sttp.openai.OpenAIExceptions$OpenAIException$DeserializationOpenAIException: expected string got dictionary


dgolda:
sttop-openai version 0.3.2

While calling listRunSteps, I encountered the following exception:
```
sttp.openai.OpenAIExceptions$OpenAIException$DeserializationOpenAIException: expected string got dictionary at index 2990
	at sttp.openai.OpenAIExceptions$OpenAIException$DeserializationOpenAIException$.apply(OpenAIExceptions.scala:24)
	at sttp.openai.OpenAIExceptions$OpenAIException$DeserializationOpenAIException$.apply(OpenAIExceptions.scala:27)
	at sttp.openai.json.SttpUpickleApiExtension$.$anonfun$deserializeJsonSnake$1(SttpUpickleApiExtension.scala:47)
	at sttp.openai.json.SttpUpickleApiExtension$.$anonfun$deserializeRightWithMappedExceptions$1(SttpUpickleApiExtension.scala:39)
	at sttp.client4.MappedResponseAs.$anonfun$mapWithMetadata$1(ResponseAs.scala:462)
	at sttp.client4.internal.BodyFromResponseAs.$anonfun$doApply$4(BodyFromResponseAs.scala:26)
	at sttp.monad.IdentityMonad$.eval(MonadError.scala:221)
	at sttp.client4.internal.BodyFromResponseAs.$anonfun$doApply$3(BodyFromResponseAs.scala:26)
	at sttp.monad.syntax$MonadErrorOps.map(MonadError.scala:71)
	at sttp.client4.internal.BodyFromResponseAs.$anonfun$doApply$2(BodyFromResponseAs.scala:26)
	at sttp.monad.IdentityMonad$.flatMap(MonadError.scala:216)
	at sttp.monad.syntax$MonadErrorOps.flatMap(MonadError.scala:72)
	at sttp.client4.internal.BodyFromResponseAs.doApply(BodyFromResponseAs.scala:25)
	at sttp.client4.internal.BodyFromResponseAs.$anonfun$apply$1(BodyFromResponseAs.scala:16)
	at sttp.monad.syntax$MonadErrorOps.map(MonadError.scala:71)
	at sttp.client4.internal.BodyFromResponseAs.apply(BodyFromResponseAs.scala:16)
	at sttp.client4.internal.httpclient.BodyFromHttpClient.apply(BodyFromHttpClient.scala:20)
[...]

PR #322, opened 2025-04-14
Fix FileSearch format in FileSearchToolCall (#321)

dgolda:
I've changed
```
fileSearch: Map[String, String]
```
to
```
fileSearch: Option[FileSearch] = None
```

and defined FileSearch class with fields as in OpenAi documentation.

It works for me.


adamw:
Great, thanks!

PR #340, opened 2025-05-21
Add "strict" structured-output support for function calling

pstutz:
Tested it with the OpenAI API: `sbt "examples3/runMain examples.StrictStructuredOutputExample"`, seems to work.

Background on the feature: https://openai.com/index/introducing-structured-outputs-in-the-api/


adamw:
Thanks!

PR #374, opened 2025-07-31
Expose optional Usage field in completions streaming

KadekM:
https://platform.openai.com/docs/api-reference/chat_streaming/streaming#chat_streaming/streaming-usage


pielas:
LGTM! Thanks :)

PR #396, opened 2025-09-08
Support js and native in core

majk-p:


majk-p:
There seem to be too blockers now:
* [tapir-apispec-docs](https://github.com/softwaremill/tapir/blob/master/build.sbt#L1135) are not cross-compiled, although all its dependencies seem to be native-ready
* [ujson-circe](https://github.com/com-lihaoyi/upickle/blob/main/build.mill#L222) is not cross-compiled, it seems to only depend on Circe which should be native 0.5.x ready, but I didn't have a close look

adamw:
Ah :) Yeah we should be able to take care of tapir-apispec-docs quickly, if the dependencies are there.

majk-p:
So the good part is that https://github.com/com-lihaoyi/upickle/pull/680 is merged, bad side of it is that I somehow forgot to add JS so it's only native for now. When tapir-apispec-docs are ready we can start with SN support and contribute JS later.

PR #461, opened 2026-04-09
feat: Miscellaneous changes aimed at providing better compatibility with models hosted on vertex AI

wookievx:
This PR delivers changes that provide compatibility with google vertex AI hosted anthropic/chat-completion compatible models (like Gemini in the case of the latter).


wookievx:
@adamw I verified locally that with this PR both anthropic message API and completions API work when served by vertex AI. Probably will also deliver something like this for the response API. I think is it valuable thing to have in sttp-ai to support it.

adamw:
Thanks :) Looking forward for more changes

We might end up with everything being optional ;-) That's fine of course if it's the reality, just a side-comment

PR #466, opened 2026-05-05
Support for structured outputs in agent loop and improvements in clients

### LINKED ISSUE #436: Add support for structured outputs in Agent Loop
Currently, agent loops (`AgentResult[T]`) always return `String` for `finalAnswer`, even though the generic OpenAI API already supports structured outputs via `ResponseFormat.JsonSchema`. This means developers must manually parse agent outputs, losing type safety and compile-time guarantees. 

The proposal is to extend agent loops to support structured outputs in the same way as regular API calls, allowing `AgentConfig[T]` to specify a response schema and return typed results like `AgentResult[WeatherSummary]` instead of `AgentResult[String]`.

kubinio123:
DONE: 
* Support for structured output in agent core:
   * Add `ResponseSchema[T]` for structured responses instead of plain text
   * `Agent.runAs[T]` returning `AgentResult[Either[AgentParseError, T]]`
   * Update the system prompt to instruct the model to call finish with a JSON payload matching the response schema
* Utilities in clients:
   * Typed chat helpers for using structured output (tapir schemas + upickle)
   
I think this solves the https://github.com/softwaremill/sttp-ai/issues/436.

However https://github.com/softwaremill/sttp-ai/issues/206 is still a valid idea to explore; less explicit approach relying solely on meta programming. I think that can still be a subject to some POC 🤔


core/src/main/scala/sttp/ai/core/agent/AgentConfig.scala
```diff
@@ -30,7 +36,12 @@ object AgentConfig {
     }
   }
 
-  private def buildSystemPrompt(maxIterations: Int): String =
+  private def buildSystemPrompt(maxIterations: Int, structuredFinish: Boolean): String = {
+    val structuredAddendum =
+      if (structuredFinish)
+        "\n\nWhen calling the finish tool, you MUST pass a JSON object matching the provided input schema for the final answer."
```
adamw: shouldn't this be - the provided *output* schema?

examples/src/main/scala/examples/AgentLoopExample.scala
```diff
@@ -74,3 +74,51 @@ object AgentLoopExample extends App {
     }
   finally backend.close()
 }
+
+object TypedAgentLoopExample extends App {
```
adamw: maybe this should be a separate example file

adamw (approved):

kubinio123:
Thanks, pushed the updates ✔️

PR #472, opened 2026-05-18
feat: Simplified and properly bi-directional message read-writer

wookievx:
Allowing the message model to be used for both encoding and decoding chat-completion requests.


adamw:
Thanks :)

PR #479, opened 2026-06-05
Agent loop: Finish when the agent's response contains a response with no tool calls

korlowski:
Using a tool to return the answer to the user is not the recommended approach. With this change, the agent's text response when no tools are used will be used as the final answer. For typed `runAs[T]` loop, the structured JSON schema-based response is expected from the agent. 

Additionally:
- Since the finish tool was the only "system" tool in the codebase, I removed the system tools feature for simplicity. This also simplified the AgentConfig creation.
- Adjusted some tests that were failing due to imprecise prompts and non-strict JSON schema for tool use (e.g. OpenAI was making up calculator operations that weren't specified in the schema)
- I removed the prompt reminder that the agent should reply with the given schema. We specify the response schema in the API request and the agents don't need to be reminded of the OpenAI/Claude API specification.


adamw:
Good call with simplifying the tool model :)

Preliminary review from Claude:

```
  Overall assessment: A clean, well-motivated simplification. The loop stays a pure tail-recursive fold over immutable history with no shared mutable state. Native structured output is the right call. There is one real correctness gap that all three reviewers
  independently flagged, plus minor consistency nits.

  ---
  🔴 Must fix — termination condition misses the empty-response case

  core/src/main/scala/sttp/ai/core/agent/Agent.scala:40

  if (response.toolCalls.isEmpty && response.textContent.nonEmpty)

  The condition is a conjunction. When the model returns no tool calls AND empty text, it falls into the else (tool-execution) branch, maps over an empty tool-call seq (no-op), appends an empty assistant turn, and recurses — burning every remaining iteration on
  no-progress API round-trips, then returning "No answer available" with finishReason = MaxIterations.

  - Empirically reproduced by the functionality reviewer: AgentResponse("", Seq.empty, EndTurn) with maxIterations=4 → 4 paid round-trips, useless answer, instead of terminating on iteration 1.
  - This is a realistic model output (content filter / refusal / reasoning-only with empty visible content).
  - It also contradicts the new system prompt, which promises "A response that contains no tool calls terminates the agent loop" — unconditionally.
  core/src/main/scala/sttp/ai/core/agent/Agent.scala:40
  if (response.toolCalls.isEmpty && response.textContent.nonEmpty)

  The condition is a conjunction. When the model returns no tool calls AND empty text, it falls into the else (tool-execution) branch, maps over an empty tool-call seq (no-op), appends an empty assistant turn, and recurses — burning every remaining iteration on
  no-progress API round-trips, then returning "No answer available" with finishReason = MaxIterations.

  - Empirically reproduced by the functionality reviewer: AgentResponse("", Seq.empty, EndTurn) with maxIterations=4 → 4 paid round-trips, useless answer, instead of terminating on iteration 1.
  - This is a realistic model output (content filter / refusal / reasoning-only with empty visible content).

  core/src/main/scala/sttp/ai/core/agent/Agent.scala:40

  if (response.toolCalls.isEmpty && response.textContent.nonEmpty)

  The condition is a conjunction. When the model returns no tool calls AND empty text, it falls into the else (tool-execution) branch, maps over an empty tool-call seq (no-op), appends an empty assistant turn, and recurses — burning every
  remaining iteration on no-progress API round-trips, then returning "No answer available" with finishReason = MaxIterations.

  - Empirically reproduced by the functionality reviewer: AgentResponse("", Seq.empty, EndTurn) with maxIterations=4 → 4 paid round-trips, useless answer, instead of terminating on iteration 1.
  - This is a realistic model output (content filter / refusal / reasoning-only with empty visible content).
  - It also contradicts the new system prompt, which promises "A response that contains no tool calls terminates the agent loop" — unconditionally.

  Recommended fix: terminate whenever toolCalls.isEmpty, using whatever text is present (possibly empty) as the final answer:

  if (response.toolCalls.isEmpty)
    monad.unit(AgentResult(response.textContent, iteration + 1, toolCallRecords, FinishReason.NaturalStop))

  This branch is also untested — see test gap below.

  🟡 Suggestions

  1. Inconsistent responseSchema default across backends. Claude's backend declares responseSchema: Option[ResponseSchema[_]] = None (ClaudeAgent.scala:20); OpenAI's has no default (OpenAIAgent.scala:18). Both are private backends always
  constructed with all args explicitly, so the default is dead. Drop it from Claude for symmetry.
  2. Claude outputConfig always sent, even when empty. ClaudeAgent.scala:26-27,109 builds OutputConfig(format = responseSchema.map(...)) and always passes Some(outputConfig). When there's no schema this serializes to "output_config": {} on
  every request — verified benign (doesn't trip structured-output handling), but asymmetric with OpenAI's clean Option[ResponseFormat]. Mirror the OpenAI pattern:
  private val outputConfig: Option[OutputConfig] =
    responseSchema.map(rs => OutputConfig(format = Some(OutputFormat.JsonSchema(rs.schema))))
  3. OpenAI strict = Some(true) now forced on every user tool (OpenAIAgent.scala:42, was false). Verified safe for tools built via AgentTool.fromFunction (its schema path already emits additionalProperties:false, all-required, nullable
  options). Residual low risk only for tools built from a hand-supplied schema (optionals get forced into required but not made nullable). Worth a doc note that custom tool schemas must be strict-compatible.
  4. AgentConfig private constructor + apply factory is now only justified by defaulting systemPrompt. That's a legitimate (if thin) smart-constructor reason, so it's defensible — but note buildSystemPrompt bakes in maxIterations at
  construction, so a later .copy(maxIterations = …) yields a stale prompt. Minor.

  🟢 Test coverage

  - HIGH — add a test for the empty-response edge (the one genuinely-new, untested branch): AgentResponse("", Seq.empty, EndTurn) with maxIterations=5, asserting current behavior. This pins the conjunction and would catch a regression if
  someone simplifies && away. (If you adopt the Must-fix above, update the assertion to NaturalStop/iterations=1 accordingly.)
  - LOW — no unit test for the new schema→request translation in either backend (OpenAI ResponseFormat + strict=true; Claude OutputConfig). Currently only exercised by paid, skippable integration tests. A small per-backend unit test
  asserting the outgoing request carries the schema is the highest-value coverage after the edge case.
  - LOW — duplicate tests: "handle tool execution errors gracefully" and "send RuntimeException to LLM and continue loop" now assert the same path (RuntimeException → recover → continue → NaturalStop); consider merging.
  - Deletions are correct: finish-tool / systemTools / reserved-name tests are genuinely obsolete, no real scenario lost.
  - Integration prompt rewordings are legitimate, not flakiness-masking — they align prompts with assertions and one even tightens finishReason from (ToolFinish or NaturalStop) to shouldBe NaturalStop. Worth confirming the conditional-logic
  reword ("tell me the result of the calculation as well as the weather") was an intentional assertion-alignment.
```

adamw:
Re-review:

```
  Headline: the previous must-fix is resolved ✅

  Agent.scala:40 now terminates unconditionally on response.toolCalls.isEmpty (returning response.textContent with NaturalStop), and there's a new test pinning the empty-text case (AgentResponse("", Seq.empty, …) → NaturalStop, iterations = 1). The earlier conjunction bug
  (&& textContent.nonEmpty burning iterations) is gone. Prior suggestions #1 (Claude responseSchema default) and #2 (outputConfig always-Some) are also addressed — Claude's outputConfig is now Option, emitted only when a schema is set.

  Remaining findings

  1. ADR contradicts the implementation and was not updated — docs/adr/0001-agent-stop-signal.md
  The ADR is titled "Use a dedicated finish tool to terminate the agent loop" and explicitly rejects "inferring completion from a structured output format" as "fragile." This PR deletes the finish tool and implements exactly the rejected approach. The ADR is untouched in
  the diff. An ADR that documents the opposite of the code is worse than none — supersede it (0002) or mark it Status: Superseded.

  2. stopReason is computed by both backends but never consumed; truncated answers are mislabeled — Agent.scala:40-49
  The loop branches only on toolCalls.isEmpty and always returns NaturalStop. Both backends fully compute response.stopReason (incl. StopReason.MaxTokens) and the result is discarded. Concrete failure: a final text answer cut off by the token limit (stopReason =
  MaxTokens, no tool calls) is returned as FinishReason.NaturalStop with a truncated answer — the caller can't tell it was incomplete. Relatedly, FinishReason.TokenLimit and FinishReason.Error are now unreachable dead cases (no producers anywhere). The natural fix the PR
  left undone: map MaxTokens → TokenLimit in the no-tool-call branch.

  3. responseSchema is applied on every request, including tool-calling iterations — OpenAIAgent.scala:24-31, ClaudeAgent.scala:26-27,108
  With a schema set, response_format=json_schema(strict) (OpenAI) / output_config (Claude) plus tools are sent on all iterations, not just the terminating turn. OpenAI documents this combination as supported (tool turns emit no schema-bound content), so it's likely fine
  there. For Claude the tools + output_config interaction is riskier and is exercised only by the key-gated integration spec. Worth confirming the typed integration test actually ran green against the live API; unit tests use a stub backend and can't catch it.

  4. maxIterations + runAs[T] returns Left(AgentParseError) from raw tool output — untested — Agent.scala:20-29
  When the loop exhausts iterations, finalAnswer comes from extractFinalAnswer (last assistant text / raw tool result / "No answer available"). For a typed agent that string won't parse as T. Previously the finish tool guaranteed schema-shaped JSON; nothing re-establishes
  that on the cap path. Defensible behavior, but it's a behavior change with no test.

  5. No unit test verifies the schema reaches the outgoing request — OpenAIAgentIntegrationSpec / ClaudeAgent
  The core "responseSchema round-trip" test uses a stub backend that ignores the schema entirely — it only proves a JSON string parses, identical to the runAs[T] tests. No unit test captures the serialized ChatBody/MessageRequest to assert response_format/output_config is
  present. A regression dropping the schema from the request would pass every unit test. Highest-value coverage to add after this change.

  6. Dangling README references — README.md:664, :842
  Line 664 still lists "Type-safe tool definitions with automatic finish mechanism" — the finish mechanism is gone. Line 842 lists Error as a possible finishReason, which is now unreachable (see #2).

  7. buildSystemPrompt bakes maxIterations into the prompt at construction — AgentConfig.scala:18-20 (minor, latent)
  AgentConfig is case class private but .copy is still public; config.copy(maxIterations = N) leaves systemPrompt advertising the old number. No current caller uses .copy, so latent. With validation gone, the private-constructor/factory now exists solely for this
  prompt-baking — a derived/lazy systemPrompt would be cleaner and copy-safe.

  8. Test cleanup (minor) — Two near-identical "no tool calls → NaturalStop" tests differ only in text value; the empty-text variant is the meaningful one. "accept valid user tools" now just asserts a Seq round-trips (its negative companion, the reserved-name test, was
  correctly deleted) and tests no agent logic.
```

Probably worth fixing, at least some of these :)

korlowski:
1. Updated - please check if the reasoning makes sense
2. Added a case for MaxTokens 
3. Except the last iteration, we don't know in advance whether a tool will be called or not... This makes me think that on the last iteration, we should pass empty `tools` so as not to tempt the agent to call them (the prompt instructs it not to but who knows if the agent will respect it). But that's a separate change
4. It's a good point but independent of my changes so we can fix it as a next step. I don't like this `AgentResult[Either[AgentParseError, T]` response precisely because it doesn't capture these cases well. Maybe we can turn it to `AgentResult[Option[T]]` and return parsing error differently
5. Maybe it's enough that integration tests cover this case. No unit test was checking the details of the json that goes out in the request.
6. Updated
7. Minor & preexistent edge case, left it for now.
8. Updated

adamw:
Thank you!

PR #482, opened 2026-06-10
BeforeToolCall and afterToolCall hooks in the agent loop

korlowski:
Configurable and effectful `beforeToolCall` and `afterToolCall` hooks. These can be used for example for logging. This change pushes the `F` parameter into new places.

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,6 +677,7 @@ import sttp.ai.core.json.SnakePickle
 import sttp.ai.openai.OpenAI
 import sttp.ai.openai.agent.OpenAIAgent
 import sttp.client4.DefaultSyncBackend
+import sttp.shared.Identity
 import sttp.tapir.Schema
 
 object BasicExample extends App {
@@ -689,7 +690,7 @@ object BasicExample extends App {
     s"The weather in ${input.location} is 22°C, sunny"
   }
 
-  val config = AgentConfig(
+  val config = AgentConfig[Identity](
     maxIterations = 5,
     userTools = Seq(weatherTool)
   )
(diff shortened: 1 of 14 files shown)


adamw:
Claude review:

```
  Correctness — clean

  - The synchronous→effectful refactor faithfully preserves error-handling semantics. Old catch { case e: Exception } maps exactly to handleError { case e: Exception => ... } (non-Exception Throwables still re-raise via the monad), and throw ex → monad.error(ex) (still
  throws synchronously under Identity).
  - Hooks fire exactly once per tool call, in before→after order, across all paths: success, parse failure, tool-not-found. The new test (AgentSpec.scala:611) asserts this ordering with inOrderOnly.
  - monad.eval correctly defers the parse and tool.execute side effects for lazy F (IO/Task) — an improvement over the old eager evaluation.
  - Unit tests pass, full cross-module compile succeeds (functionality reviewer verified).

  Main concern — API ergonomics (simplicity reviewer)

  Parameterizing AgentConfig[F[_]] imposes a large, permanent tax for a feature most users won't use. Because F only appears in defaulted-None parameters, inference can't recover it, so every call site must now annotate AgentConfig[Identity] / [IO] / [Task] — 9 sites in
  the README, both examples, ~15 in tests, both integration specs. Tools themselves are synchronous (AgentTool.execute: T => String); the only thing forcing F onto the config is the two hooks.

  Suggested alternative: move the hooks onto Agent / OpenAIAgent.apply / ClaudeAgent.apply, which already carry [F] and a MonadError[F]. Hooks are loop-execution concerns, not declarative config. This delivers the identical capability while leaving AgentConfig
  unparameterized — eliminating essentially all the README/examples/test churn in this diff. Worth raising with the author before merge.

  Minor points

  - Stack safety (low risk): runToolCalls (Agent.scala:294) is newly recursive via flatMap, eager under Identity. Depth = tool calls in one response, normally small, so not a practical concern — but a one-line // not stack-safe; relies on small tool-call counts comment
  would help. loop was already recursive, so no regression there.
  - Hook failures abort the whole run, discarding accumulated toolCallRecords. This is documented in the README ("an error in either hook interrupts the run") so it's intentional — just confirm that's the desired contract.
  - Unrelated churn: inlining extractFinalAnswer and switching monad.flatMap(response){…} to response.flatMap{…} are stylistic and enlarge the diff. Harmless but unnecessary.
  - Mutable ListBuffer in the new test is fine — single-threaded, test-local, the only practical way to observe Identity hook ordering.
```

Regarding the ergonomics, I can see the point - and adding explicit type-parameters a common pain. Although I'm not sure if we won't end up with parametrising the config anyway when we add MCP tools. @kubinio123 I think you sketched how this might look like? But moving this to `Agent` is also something to consider. Or introducing a separate class, keeping `AgentConfig` for general configuration and having `AgentEnvironment` or `AgentCallbacks[F]` separately

kubinio123:
The `AgentTool.execute` will need to be effectful for sure when integrating MCP tools. That would mean that `tools` in `AgentConfig` will also need to be effectful. I think we can parametrize `AgentConfig` now. In the long term I would vote for creating more of a fluent API for agent creation, but that's my personal preference in libraries in general, and also what tapir or sttp does for example. I am going in that direction in chimp server side too.

korlowski:
Added API for building the agent. Also used it to clean up some bloated code in the AgentSpec

PR #486, opened 2026-06-24
feat: Exposing cache control API in OpenAI and Anthropic API's

wookievx:
Adding support for OpenAI chat-completions caching related request parameters and Anthropic create message cache_control parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ object Main:
     response.body match {
       case Right(messageResponse) =>
         messageResponse.content.foreach {
-          case ContentBlock.Text(text, _) => println(text)
+          case ContentBlock.Text(text, _, _) => println(text)
           case _ => // Handle other content types if needed
         }
         println(s"Usage: ${messageResponse.usage}")
@@ -292,6 +292,8 @@ val request = MessageRequest.simple(
 #### Advanced Parameters
 
 ```scala
+import sttp.ai.claude.models.CacheControl
+
 val request = MessageRequest(
   model = "claude-3-sonnet-20240229",
   messages = messages,
@@ -301,9 +303,26 @@ val request = MessageRequest(
   topK = Some(40),                   // Top-k sampling
   stopSequences = Some(List("\n\n")), // Stop generation at sequences
   system = Some("Be concise and helpful."),
-  tools = Some(tools)                // Tool calling support
+  tools = Some(tools),                // Tool calling support
+  cacheControl = Some(CacheControl.Ephemeral())  // Optional cache control
 )
 ```
+
+Regarding caching and usage, it is important to highlight model and formula used to calculate the number of input tokens 
+consumed by the model (relevant for billing and context window management): 
+
+```scala
+case class Usage(
+    inputTokens: Int,
+    outputTokens: Int,
+    cacheReadInputTokens: Option[Int] = None,
+    cacheCreationInputTokens: Option[Int] = None
+) {
+  def totalInputTokens: Int = inputTokens + cacheReadInputTokens.getOrElse(0) + cacheCreationInputTokens.getOrElse(0)
+  def totalTokens: Int = totalInputTokens + outputTokens
+}
+```
+This is a breaking change compared to old version of this library (which ignored cache tokens).
 
 ### Claude Structured Outputs
 
@@ -401,7 +420,7 @@ object StructuredOutputExample:
     response.body match {
       case Right(messageResponse) =>
         messageResponse.content.foreach {
-          case ContentBlock.Text(text, _) =>
+          case ContentBlock.Text(text, _, _) =>
             println("Structured JSON output:")
             println(text)
 
--- a/claude/src/main/scala-2/sttp/ai/claude/json/ClaudeDerivedCodecs.scala
+++ b/claude/src/main/scala-2/sttp/ai/claude/json/ClaudeDerivedCodecs.scala
@@ -27,6 +27,7 @@ object ClaudeDerivedCodecs {
   implicit val citationsConfigCodec: Codec[CitationsConfig] = deriveConfiguredCodec
   implicit val webSearchResultCodec: Codec[WebSearchResult] = deriveConfiguredCodec
   implicit val citationCodec: Codec[Citation] = deriveConfiguredCodec
+  implicit val cacheControlCodec: Codec[CacheControl] = deriveConfiguredCodec
 
   implicit val imageSourceCodec: Codec[ImageSource] = deriveConfiguredCodec
   implicit val documentSourceCodec: Codec[DocumentSource] = deriveConfiguredCodec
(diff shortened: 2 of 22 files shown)


adamw:
## Review: Exposing cache control API in OpenAI and Anthropic APIs

**Overall:** Mostly correct. I verified the wire shapes against the Anthropic and OpenAI docs: the request-level `cache_control` (Anthropic supports a top-level auto-placement param), the content-block `cache_control` placements, the `{"type":"ephemeral","ttl":...}` JSON shape, and the OpenAI `"24h"` / `"in_memory"` values are all right. Notably `"in_memory"` (underscore) matches what the OpenAI API actually expects — see [openai-python#2883](https://github.com/openai/openai-python/issues/2883), where the SDK type wrongly used a hyphen. The one real bug is tool-level cache-control placement.

### Findings (most-severe first)

1. **`claude/src/main/scala/sttp/ai/claude/models/Tool.scala` — tool `cache_control` is nested inside `input_schema` instead of placed on the tool object, so tool prompt-caching silently never takes effect.** `cacheControl` was added to `ToolInputSchema`, and `Tool.customRW` serializes it under `input_schema` → `{"name":...,"description":...,"input_schema":{"type":"object",...,"cache_control":{"type":"ephemeral"}}}`. The Anthropic API expects `cache_control` as a **sibling** of `input_schema` on the tool object, not inside the JSON schema. A user who sets it to create a tool cache breakpoint gets no caching (the nested key is ignored as an unknown schema field). The round-trip unit test passes because uPickle reads back whatever it writes — it doesn't validate against the real API shape.

2. **`README.md` (Advanced Parameters example, ~line 290) — the example no longer compiles: missing comma before the new `cacheControl` line.** The diff inserts `cacheControl = Some(CacheControl.Ephemeral())` directly after `tools = Some(tools)` with no trailing comma on the prior line, producing two named arguments with no separator. (`CacheControl` is also unimported in that snippet.) It's a plain ` ```scala ` block, not `mdoc:compile-only`, so CI won't catch it — but anyone copying the example hits a syntax error.

3. **`claude/src/main/scala/sttp/ai/claude/models/ContentBlock.scala` — `WebSearchResult.cacheControl` is dead API surface.** `WebSearchResult` is a response-only sub-result (returned inside `WebSearchToolResultContent`), never sent in a request, so a `cache_control` field on it can never do anything. Minor, but it widens the public API with a meaningless option and invites confusion about where caching is actually configurable.

4. **`claude/src/main/scala/sttp/ai/claude/models/Usage.scala` — `totalTokens` semantics changed.** It now returns `inputTokens + cacheReadInputTokens + cacheCreationInputTokens + outputTokens` rather than `inputTokens + outputTokens`. Reasonable (and `totalInputTokens` is a useful addition), but it's a behavioral change to an existing public method — any caller relying on the old value will see different numbers once caching is in play. Worth a changelog note.

Findings 1 and 2 are worth fixing before merge; 3 and 4 are judgment calls for the author.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

wookievx:
Updated the PR for the usage of circe as json library.

Regarding the issues:
* fixed 1/2 and 3 (By mistake I placed the cache control on the wrong level).
* 4 I could add changelog note, not certain what is the preferred approach in this project. New logic is definitely the correct one, from the docs: https://platform.claude.com/docs/en/build-with-claude/prompt-caching#tracking-cache-performance, if you need to track context-window length, ignoring cached tokens will lead to problems.

wookievx:
@adamw If you could take a look at this PR again, it would be great.

adamw:
## Re-review: Cache control for OpenAI & Anthropic

Thanks for the updates. Since the last pass the tool `cache_control` was moved off `ToolInputSchema`, `WebSearchResult.cacheControl` was removed (finding 3), and the codecs migrated to circe. CI is green and I re-verified the wire shapes against the Anthropic docs. Most findings are resolved — but the finding 1 fix is incomplete: it still silently drops cache control for **custom** tools.

### 🔴 Blocking — `Tool.Custom` cache control is dropped on serialization

`claude/src/main/scala/sttp/ai/claude/json/ClaudeManualCodecs.scala` (the `Tool` encoder, ~L76-81)

The `Custom` branch is hand-written and only emits three fields:

```scala
case x: Tool.Custom =>
  Json.obj(
    "name" := x.name,
    "description" := x.description,
    "input_schema" := x.inputSchema
  )   // cacheControl is never written
```

So setting `cacheControl` on a custom tool produces **no cache breakpoint** — the same silent no-op as the original finding 1, just relocated from `input_schema` to the manual encoder. Note the inconsistency: `Tool.WebSearch` serializes via `x.asJson` (the derived codec), so its `cache_control` **is** emitted correctly — only `Custom` is broken.

No test catches this — the unit round-trip test only covers request-level and `ContentBlock.Text` cache control. A `Tool.Custom` round-trip would fail on `Some(...)` vs `None` and would have surfaced it.

**Fix:** add `"cache_control"` (drop-if-`None`) to the `Custom` branch, or let `Custom` serialize via its derived codec the way `WebSearch` does — and add a `Tool.Custom` round-trip test.

### ✅ Verified correct
- **Top-level request `cache_control`** (`MessageRequest.cacheControl`) — valid; Anthropic supports it for automatic caching (breakpoint on the last cacheable block).
- **`CacheControl` wire shape** — `{"type":"ephemeral"}` / `{"type":"ephemeral","ttl":"1h"}` via the derived codec + snake-case discriminator config.
- **Content-block cache control** (`Text`/`Image`/`ToolResult`/`Document`) — emitted as a sibling of `type`, correct placement.
- **OpenAI** `promptCacheKey` / `promptCacheRetention`, `"in_memory"` / `"24h"` — correct and consistent with the existing `ResponsesRequestBody` fields.

### 🟡 Minor (non-blocking)
- **Codec declaration order** — `cacheControlCodec` is declared *after* the codecs that consume it (`contentBlockCodec`, `messageRequestCodec`) in both `ClaudeDerivedCodecs` files. CI green ⇒ no runtime NPE, but it breaks the file's otherwise dependency-ordered layout; consider moving it up with the other model codecs.
- **README Advanced-Parameters snippet** — the missing comma is fixed, but `CacheControl` isn't imported in that plain (non-`mdoc`) block, so a copy-paste won't compile.
- **`Readme-prefix.md` fixture (655 lines)** duplicates the README and will drift; a smaller synthetic large-text block would be lower-maintenance. The integration assertion also depends on the doc exceeding the model's ~1024-token minimum cacheable size — fine today, but mildly flaky-prone (opt-in test, so non-blocking).
- **Changelog note** for the `Usage.totalTokens` semantics change — worth adding, as you offered.

**Bottom line:** one blocking fix (Custom-tool encoder + test); everything else is clean or minor.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

adamw:
@wookievx Sorry I was on vacations. I posted a re-review :)

flsh86 (approved):
LGTM

PR #491, opened 2026-06-25
Make AgentTool effectful

korlowski:
AgentTool.run and hooks are already effectful so it only makes sense for AgentTool.execute to be


adamw:
Thanks!

PR #513, opened 2026-07-15
Mcp tools integration

flsh86:
Load agent tools from MCP servers via chimp client

## What

New Scala-3-only mcp module: agents can load their tools from MCP servers instead of hand-defining each one with AgentTool.fromFunction.

val transport = ClientStdioTransport(List("npx", "-y", "my-mcp-server"))
val client = McpClient[Identity](transport, Implementation("my-agent", "1.0"))
val mcpTools = McpTools.fromClient(client)          // discovery + schema conversion
val agent = ClaudeAgent.synchronous(...).tools(mcpTools).build

## How

- Built on chimp 0.4.0's McpClient[F] — same sttp MonadError[F] abstraction as the agent loop, so all chimp transports work (stdio/HTTP/streaming). No changes to core: the integration is glue over the existing AgentTool.dynamicF.
- McpTools.fromClient drains tools/list pagination, decodes schemas into sttp.apispec.Schema (fail-fast McpToolConversionException on invalid ones), executes via tools/call, and renders results as text (server-side errors go back to the LLM as Tool execution failed: ...).
- Optional namePrefix avoids name collisions across servers/manual tools; the caller owns the client lifecycle; the tool list is a snapshot (no tools/list_changed).

## Testing

- Unit tests (stub client): pagination, prefixing, description fallback, schema failure, result rendering.
- Wire-level test: in-process chimp MCP server over HTTP on every sbt test run — no API keys/containers.
- Docs: compile-checked mdoc page docs/agents/mcp.md.
- Manually smoke-tested end-to-end: ClaudeAgent + @modelcontextprotocol/server-everything over stdio against the live Claude API — agent discovered 13 tools, called get-sum through the prefix mapping, correct result.

## Follow-up PRs needed

1. OpenAI strict-mode compatibility (main blocker for arbitrary servers): the OpenAI backend sends strict: true, which rejects MCP schemas lacking additionalProperties: false — normalize schemas or make strict configurable. Also: Claude backend flattens nested schemas. Both currently documented as caveats only.
2. Robustness/convenience: MCP tool-name validation, duplicate-name detection, optional fromMcpServer(...)-style shortcuts.


adamw:
Quick review by Claude:

PR #513 review — MCP tools integration

Verdict: solid, well-tested PR — approvable with minor changes. The new mcp module (~80 lines of main code + thorough tests) cleanly glues chimp 0.4.0's McpClient[F] onto the existing AgentTool.dynamicF, with no changes to core. All five reviewers (correctness, structure, Scala/FP, tests, simplicity) independently confirmed the fundamentals: pagination logic, exception capture in F's error channel, prefix-name mapping (LLM sees calc_add, server gets add), test scoping in build.sbt, and the docs example all check out against the real core agent code.

Main discussion point (major, low likelihood)

One bad tool schema disables the whole server's tool set (McpTools.scala, fromClient/toAgentTool). All tools are converted inside a single monad.eval, and the first schema that fails to decode throws McpToolConversionException, so the caller gets zero tools instead of the N−1 good ones. That contradicts the docs' own advice to "wrap the problematic tool manually with AgentTool.fromFunction" — you can't reach the good tools to combine them. The PR describes fail-fast as intentional, so this may be a deliberate trade-off, but consider per-tool decoding that skips (or reports) failures without discarding the successes.

Worth fixing before merge (small)

- Wire test runs in the default sbt test loop — McpToolsHttpServerSpec binds a real Netty server on every test run, but doesn't follow the repo's *IntegrationSpec/integration package convention for tests excluded from the fast unit loop. It's a valuable test (proves real wire encoding decodes through inputSchema.as[Schema]); just relocate/rename or tag it.
- Simplification: flatMap { … monad.eval(…) } in fromClient collapses to a single .map(_.map(toAgentTool(client, _, namePrefix))) — sttp's MonadError.map captures a throwing function identically, so behavior is unchanged.
- One missing test: precedence when both content blocks and structuredContent are present (blocks win, structuredContent silently dropped) — no test pins this contract.
- Scope hygiene: the .gitignore additions and deletion of two docs/superpowers planning files are unrelated to MCP — better split out, or at least mentioned in the PR description.

Minor / nits

- listAllTools recursion isn't stack-safe on Identity/Try and has no guard against a server that loops the same nextCursor forever — a page cap or visited-cursor check would be defensive. Low real-world risk.
- Duplicate exposed tool names silently collapse in the agent's toolMap (last one wins) — namePrefix only disambiguates across servers, not within one. Already flagged as a follow-up in the PR description.
- A successful empty result renders as "" to the LLM; a sentinel like "(empty result)" would read better.
- case ToolContent.Text(_, text) is positional — case t: ToolContent.Text => t.text survives chimp field reordering.
- The wire test uses ox.supervised via an undeclared transitive dependency (through tapir-netty-server-sync); an explicit Test-scoped ox dep would harden it.
- The test->test edge on core is unused by the mcp tests — though it matches the repo-wide convention, so keeping it is defensible.

Explicitly checked and fine

Package naming sttp.ai.core.agent.mcp follows the streaming-module convention with no split-package risk; the separate Scala-3-only module is forced (chimp is Scala 3 only, core cross-builds 2.13); docs depending on mcp is required for the mdoc snippet; throw-inside-monad.eval mirrors the exact pattern Agent.executeTool already uses; transport failures correctly reach the agent's ExceptionHandler on all effect types; renderResult is pure with every branch tested; the unit suite is minimal and orthogonal with strong assertions (e.g. cursor and call recording proving pagination and prefix behavior); namePrefix, the description→title→name fallback, and the docs page's backend-caveats section all earn their place.

PR #516, opened 2026-07-16
Mcp backend schema compat

flsh86:
## What

OpenAI — strict mode made correct, and optional.
- Optional parameters survive strict mode: originally-optional properties are encoded per OpenAI's documented pattern ("type": [..., "null"], null appended to enums), recursively per nesting level, instead of being silently forced mandatory (the model previously had to invent values for them). Already-nullable anyOf properties (tapir's Option[caseclass] shape) are not double-wrapped.
- Normalization is applied only where strict is actually requested: it moved out of the shared implicit Codec[Schema] into an explicit SchemaSupport.normalizeForStrict, so structured outputs and withSchema/withTapirSchema with strict = false/absent now get a faithful encoding (previously they were rewritten unconditionally, with no opt-out).
- Escape hatch: strictTools flag on the OpenAIAgent factories (default true): OpenAIAgent.synchronous(OpenAI.fromEnv, "gpt-4o-mini", strictTools = false) registers tools non-strict with their original schemas.
- Normalizer bug fixes: a parameter literally named properties/type/required no longer corrupts the schema (the properties container was folded as if it were itself a schema — pre-existing, made destructive by the nullability rewrite, now fixed properly).

Claude — flattening removed, full pass-through.
- New additive Tool.CustomRaw(name, description, inputSchema: Json, cacheControl) (existing Tool.Custom/ToolInputSchema/PropertySchema untouched; decoder falls back to CustomRaw only for JSON Custom can't represent, so existing decode behavior is preserved).
- ClaudeAgentBackend sends the schema structure verbatim instead of collapsing it to a flat type/description/enum map, adding only a top-level "type": "object" when the schema omits it (Anthropic requires it; MCP allows {} for no-arg tools — previously a request-time 400).

Cross-cutting — the schema you get is the schema the server sent.
- AgentTool.rawJsonSchema: Json (core, default derived from jsonSchema): MCP tools carry the server's original schema JSON end-to-end, eliminating the lossy apispec round-trip (which stringified enum nulls and dropped unknown keywords).
- Both clients' send-time deepDropNullValues now exempts schema subtrees, so legitimate JSON nulls in schemas (enum: [..., null], default: null) reach the wire intact.
- Client-introduced nulls are cleaned up before they hit MCP servers: arguments that are null for parameters the server declared optional-by-absence are stripped before tools/call (validating servers, e.g. zod .optional(), reject explicit nulls); nulls for server-required params are forwarded as-is.

## Testing

- Unit/golden coverage for every rule: SchemaSupportSpec (nullability incl. enums, idempotency, anyOf guard, keyword-named params, recursion, discriminator skip, faithful codec), OpenAIAgentBackendSpec (both strict modes), ToolCustomRawSpec (wire JSON + decode fallback), ClaudeAgentBackendSpec (nested + byte-identical pass-through + type default), McpToolsSpec (raw pass-through, null-strip keep/drop), serialization specs proving schema nulls survive request encoding while other nulls are still dropped.
- API-key-gated McpAgentIntegrationSpec (self-cancels without keys — zero CI cost): an agent per backend drives a tool from an in-process chimp MCP server whose input has an optional nested object and an optional scalar. Both backends verified live on the final commit — notably proving OpenAI strict mode accepts type: ["object","null"] and chimp's format/title/$schema keywords.
- Independently verified by a wire-capture harness (fake LLM endpoints + body-logging proxy in front of a real MCP server): 25/25 checks, including the null-strip observed on the actual tools/call payload.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -161,7 +161,7 @@ lazy val mcp = (projectMatrix in file("mcp"))
       Libraries.logbackTest
     )
   )
-  .dependsOn(core % "compile->compile;test->test")
+  .dependsOn(core % "compile->compile;test->test", openai % "test->compile", claude % "test->compile")
 
 lazy val examples = (projectMatrix in file("examples"))
   .jvmPlatform(
--- a/claude/src/main/scala/sttp/ai/claude/ClaudeClient.scala
+++ b/claude/src/main/scala/sttp/ai/claude/ClaudeClient.scala
@@ -168,7 +168,7 @@ class ClaudeClientImpl(config: ClaudeConfig) extends ClaudeClient with ResponseH
       requestJson.asObject
         .flatMap(_("tools"))
         .flatMap(_.asArray)
-        .map(_.map(_.asObject.flatMap(_("input_schema"))))
+        .map(_.map(_.asObject.flatMap(_("input_schema")).filter(!_.isNull)))
 
     val cleaned = requestJson.deepDropNullValues
 
--- a/claude/src/main/scala/sttp/ai/claude/ClaudeClient.scala
+++ b/claude/src/main/scala/sttp/ai/claude/ClaudeClient.scala
@@ -9,7 +9,7 @@ import sttp.ai.core.http.ResponseHandlers
 import sttp.capabilities.Streams
 import sttp.client4._
 import sttp.model.{ResponseMetadata, Uri}
-import io.circe.Decoder
+import io.circe.{Decoder, Json}
 import io.circe.parser.decode
 import io.circe.syntax._
 import sttp.ai.claude.json.ClaudeDerivedCodecs._
@@ -117,7 +117,7 @@ class ClaudeClientImpl(config: ClaudeConfig) extends ClaudeClient with ResponseH
   override def createMessage(request: MessageRequest): Request[Either[ClaudeException, MessageResponse]] =
     claudeAuthRequestForMessage(request)
       .post(claudeUris.Messages)
-      .body(request.asJson.deepDropNullValues.noSpaces)
+      .body(serializeMessageRequest(request))
       .response(asJson_parseErrors[MessageResponse])
 
   override def listModels(): Request[Either[ClaudeException, ModelsResponse]] =
@@ -133,7 +133,7 @@ class ClaudeClientImpl(config: ClaudeConfig) extends ClaudeClient with ResponseH
 
     claudeAuthRequestForMessage(streamingRequest)
       .post(claudeUris.Messages)
-      .body(streamingRequest.asJson.deepDropNullValues.noSpaces)
+      .body(serializeMessageRequest(streamingRequest))
       .response(asStreamUnsafe_parseErrors(streams))
   }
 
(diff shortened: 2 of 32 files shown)


adamw:
## Review

### What it does

Fixes end-to-end schema fidelity for tools (especially MCP-loaded ones) across both LLM backends:

- **OpenAI**: strict-mode schema normalization now correctly makes originally-optional properties nullable (per OpenAI's documented pattern) instead of silently forcing them required, and is applied *only* when `strict = true` is actually requested — it moved out of the shared implicit `Codec[Schema]` into an explicit `SchemaSupport.normalizeForStrict`. New `strictTools` flag (default `true`) on `OpenAIAgent` factories as an escape hatch.
- **Claude**: the agent backend no longer flattens tool schemas to a `type`/`description`/`enum` map — a new `Tool.CustomRaw` carries raw JSON Schema verbatim (only adding a top-level `"type": "object"` when absent, which Anthropic requires).
- **Cross-cutting**: `AgentTool.rawJsonSchema` preserves the MCP server's original schema JSON (no lossy apispec round-trip); both clients' send-time `deepDropNullValues` now exempts schema subtrees so legitimate schema nulls (`enum: [..., null]`) reach the wire; client-introduced `null` arguments for server-optional params are stripped before `tools/call`.

### Strengths

- The problem analysis is correct and well-documented — every non-obvious decision has a comment explaining *why* (the null-drop-vs-schema-fidelity tension, the index-alignment argument for the splice-back, the `properties`-as-parameter-name corruption fix).
- Excellent test coverage: golden tests per normalization rule, both strict modes, decode-fallback behavior, serialization specs proving schema nulls survive while other nulls are still dropped, plus an API-key-gated integration spec with an in-process MCP server.
- Binary-compatibility-conscious: `strictTools` added via overloads rather than default parameters; `Tool.CustomRaw` is additive with a decode fallback that preserves historical `Tool.Custom` behavior.

### Issues

**1. High — wire regression for existing `Tool.Custom` users (Claude).** `ToolInputSchema`/`PropertySchema` use derived codecs with no None-dropping (`ClaudeDerivedCodecs.scala`), so `Tool.Custom` with `required = None` or properties with unset `description`/`enum` encodes literal nulls inside `input_schema`. Previously `deepDropNullValues` stripped them; the new splice-back in `ClaudeClient.dropNullsPreservingInputSchemas` now restores the **pre-drop** `input_schema` verbatim — so every existing `Tool.Custom` request will now send `"required": null`, `"description": null`, `"enum": null` on the wire. That's invalid JSON Schema and Anthropic may reject it. The new specs only test `CustomRaw`, so this is untested. Suggested fix: in `encodeCustomTool`, apply `deepDropNullValues` to the `Custom` case's `inputSchema.asJson` (keeping `CustomRaw` verbatim), plus a regression test for `Tool.Custom` through `createMessage`.

**2. Medium-high — Responses API not covered by null preservation (OpenAI).** `OpenAIJson.dropNullsPreservingSchemas` only exempts `tools[*].function.parameters` (chat's nested shape) and `response_format.json_schema.schema`. But the Responses API encodes function tools **flat** (`{"type":"function","name":...,"parameters":...}`) and its format under `text.format` — both paths this PR *also* adds strict normalization to (`responses/Tool.withTapirSchema`, `responsesRequestFormatEncoder`). The `null` that normalization appends to an optional property's `enum` will still be stripped at send time for Responses API requests, undermining the fix there and risking the exact strict-mode rejection this PR set out to avoid.

**3. Low-medium — `Json.Null` re-introduced by the splice-back.** Both capture sites use `.flatMap(_("parameters"))` / `_("schema")` / `_("input_schema")`, which return `Some(Json.Null)` when the key is present with a null value (e.g. chat `Tool.Function(parameters = None)`, `ResponseFormat.JsonSchema(schema = None)`). The splice-back then re-adds `"parameters": null` / `"schema": null` that were previously (correctly) dropped. A `.filter(!_.isNull)` at capture time fixes it.

**4. Low — decode fallback is narrower than documented.** The doc says schemas `Custom` can't represent fall back to `CustomRaw`, but derived decoders ignore unknown fields: a nested-object schema (`location: {type: object, properties: {...}}`) still decodes *successfully* as `Tool.Custom`, silently dropping the nested structure — so decode→re-encode round trips still corrupt nested schemas. This preserves historical behavior deliberately and doesn't affect the send path (the agent backend constructs `CustomRaw` directly), but the comment overstates the fallback's coverage.

### Nits

- `ClaudeClient.dropNullsPreservingInputSchemas` and `OpenAIJson.dropNullsPreservingSchemas` are near-duplicates; a shared "drop nulls except at these paths" helper in `core` would keep them from drifting.
- `ensureObjectType` passes non-object schemas (e.g. the boolean schema `true`, valid in JSON Schema / MCP) through unchanged — Anthropic would 400. Very edge-case.
- `isAlreadyNullableAnyOf` only matches an exact `{"type":"null"}` member; an anyOf branch with extra keys would be double-wrapped. Harmless (still valid), just not minimal.

### Verdict

The design is sound, and the strict-mode normalization and MCP pass-through logic are careful and very well tested for the paths they target. I'd hold merging until **#1** is fixed (it changes the wire format for every existing Claude `Tool.Custom` user, likely breaking them) and **#2** is either fixed or explicitly descoped (it leaves the Responses API strict path inconsistent with chat). #3 is a small, cheap fix worth bundling.

PR #520, opened 2026-07-20
Mcp tool robustness

flsh86:
## What

- Name sanitization: MCP allows tool names with dots, slashes, non-ASCII characters, and up to 128 chars; OpenAI's function calling enforces ^[a-zA-Z0-9_-]{1,64}$ and rejects the whole request otherwise — before the model even runs, and even if the offending tool is never called. McpTools sanitizes the exposed (LLM-facing) name — illegal characters become _, truncated to 64 chars after namePrefix is applied. The server always receives its own original name in tools/call; already-legal names pass through unchanged.
- Duplicate detection: two tools ending up with the same exposed name (a prefix/sanitization collision, or a server reusing one name for genuinely different tools) previously last-won silently in the agent's toolMap, routing calls meant for one tool to another. fromClient fails fast with McpToolConversionException, listing every colliding exposed name with all its original names. An MCP server re-listing the exact same tool across pages — even if its free-form _meta field varies between listings — is silently deduplicated instead of failing; a genuinely different definition under the same name still fails.
- Validation runs before schema decoding, so a collision or empty-name failure is never masked by an unrelated tool's schema-decode error.
- Effect-safe on every backend: fromClient catches with scala.util.Try and routes through monad.unit/monad.error rather than depending on a particular MonadError's map/eval to catch exceptions for it — a dependency the built-in EitherMonad doesn't satisfy (verified against sttp's actual source), and which would surface as an unrecoverable defect on ZIO backends if relied upon. A dedicated test proves the failure surfaces through the error channel even under a map/flatMap that can't catch.
- Empty-name guard: a tool whose name sanitizes to nothing (only possible when the original name is itself empty — chimp's client-side decoder doesn't validate incoming names) now fails with a clear message instead of silently producing an unusable tool.

## Testing

30 tests in the mcp module (up from 16), covering: sanitization of illegal characters, post-prefix truncation to exactly 64, original-name preservation on tools/call, exact-repeat deduplication across pages (including when _meta varies), same-name-but-conflicting-content failure, multi-way collision reporting (2-way and 3-way, with and without namePrefix), collision-before-schema-decode ordering, empty exposed names, two distinct non-ASCII names (Simplified Chinese) that sanitize to the same string, and effect-channel safety under a hand-rolled MonadError whose map/flatMap can't catch. No backend or live-path changes — the integration specs' tool names are already legal, so their behavior is untouched.

Notes

- Known, deliberately deferred scope limit: the duplicate check covers tools from a single fromClient call only. The agent's own toolMap (core.Agent) is last-wins across all merged tools (manual + every MCP source), so cross-server or MCP-vs-manual collisions aren't caught here. Closing that gap would mean adding a first-time validation to core.Agent, which every PR in this series has deliberately left untouched — and doing so would be a behavior change for the whole library (any existing app with duplicate tool names would start throwing on upgrade rather than silently degrading). namePrefix remains the recommended way to keep multiple sources distinct; left as an explicit, documented follow-up rather than fixed here.

### DIFF
--- a/docs/agents/mcp.md
+++ b/docs/agents/mcp.md
@@ -63,9 +63,12 @@ Notes:
   `fromClient` call end up with the same exposed name (a sanitization/prefix collision, or a server-side
   name reused for genuinely different tools), `fromClient` fails with a `McpToolConversionException` naming
   the colliding tools, rather than silently routing calls to the wrong one — an MCP server re-listing the
-  exact same tool across pages is harmless and is deduplicated instead. This check does not extend across
-  multiple `fromClient` calls or to manually defined tools; combining tools from different sources safely
-  is the caller's responsibility (`namePrefix` is the recommended way to keep them distinct).
+  same tool across pages is harmless and is deduplicated instead, even if its metadata varies slightly
+  between listings. This check does not extend across multiple `fromClient` calls or to manually defined
+  tools: an agent looks tools up by name, so if you combine tools from several sources without keeping
+  their exposed names distinct yourself, the last one loaded silently shadows any earlier tool with the
+  same name — no error is raised, and calls intended for one tool can silently execute another.
+  `namePrefix` is the recommended way to keep multiple sources distinct.
 * Results are rendered as text for the agent loop: text content blocks are joined with newlines, other
   block types are rendered as compact JSON, and results the server marks as errors are returned to the
   LLM prefixed with `Tool execution failed:`. Transport failures surface as exceptions and go through the
--- a/docs/agents/mcp.md
+++ b/docs/agents/mcp.md
@@ -63,8 +63,9 @@ Notes:
   `fromClient` call end up with the same exposed name (a sanitization/prefix collision, or a server-side
   name reused for genuinely different tools), `fromClient` fails with a `McpToolConversionException` naming
   the colliding tools, rather than silently routing calls to the wrong one — an MCP server re-listing the
-  same tool across pages is harmless and is deduplicated instead, even if its metadata varies slightly
-  between listings. This check does not extend across multiple `fromClient` calls or to manually defined
+  same tool across pages is harmless and is deduplicated instead, even if its free-form `_meta` field
+  varies between listings (a differing description, title, or schema still counts as a real collision).
+  This check does not extend across multiple `fromClient` calls or to manually defined
   tools: an agent looks tools up by name, so if you combine tools from several sources without keeping
   their exposed names distinct yourself, the last one loaded silently shadows any earlier tool with the
   same name — no error is raised, and calls intended for one tool can silently execute another.
--- a/docs/agents/mcp.md
+++ b/docs/agents/mcp.md
@@ -57,6 +57,11 @@ Notes:
 * `namePrefix` is optional — with `Some("mcp")`, a server tool `add` is exposed to the LLM as `mcp_add`
   (the original name is still used when calling the server). Use it to avoid name collisions with manual
   tools or tools from other MCP servers.
+* Exposed names are sanitized to the cross-backend-safe form `[A-Za-z0-9_-]`, truncated to 64 characters
+  (after prefixing) — MCP allows names with dots or slashes that OpenAI's function calling rejects. The
+  original name is still used when calling the server. If two tools end up with the same exposed name
+  (a duplicate on the server, or a sanitization/prefix collision), `fromClient` fails with a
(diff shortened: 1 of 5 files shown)


docs/agents/mcp.md
```diff
@@ -57,6 +57,19 @@ Notes:
 * `namePrefix` is optional — with `Some("mcp")`, a server tool `add` is exposed to the LLM as `mcp_add`
   (the original name is still used when calling the server). Use it to avoid name collisions with manual
   tools or tools from other MCP servers.
+* Exposed names are sanitized to the cross-backend-safe form `[A-Za-z0-9_-]`, truncated to 64 characters
```
adamw: this is a large blob of text. Some division into sections would be useful. Also making sure the text is concise & precise

adamw:
## Review

### What it does

Hardens `McpTools.fromClient` against MCP tool listings that would break or misroute agent runs:

- **Name sanitization**: exposed (LLM-facing) names are rewritten to `[A-Za-z0-9_-]` and truncated to 64 chars after `namePrefix` is applied — MCP allows dots, slashes, non-ASCII, and 128-char names, which OpenAI's function calling rejects for the whole request. The server always receives its original name in `tools/call`.
- **Duplicate detection**: two tools whose exposed names collide (via sanitization, prefixing, or a server reusing a name) now fail fast with `McpToolConversionException` listing every collision, instead of last-wins silent shadowing in the agent's tool map. Exact re-listings of the same tool across pages are deduplicated (ignoring the free-form `_meta` field).
- **Validation ordering**: name checks run before any schema decoding, so a collision isn't masked by an unrelated schema-decode error.
- **Effect safety**: failures are routed through `monad.error` via an explicit `Try`, instead of relying on the `MonadError`'s `map` to catch exceptions — which sttp's `EitherMonad` genuinely doesn't do, and ZIO only handles via its untyped defect channel.
- **Empty-name guard** plus a test-double refactor (`UnsupportedMcpClient` base trait).

### Strengths

- The effect-safety fix is correct and subtle — the accompanying `NonCatching` monad test that proves the failure surfaces through the error channel (not as an escaped exception) is exactly the right way to pin this down.
- Fail-fast on collisions is the right call: silent misrouting of tool calls is the worst failure mode, and the error lists *all* collisions with original names, deterministically sorted.
- Test coverage is excellent: 14 new tests including truncation length, original-name preservation on `tools/call`, `_meta`-varying dedup, 3-way collisions, non-ASCII names collapsing to the same string, and validation ordering.
- The docs are honest about the deliberately deferred gap (cross-source collisions in `core.Agent`'s last-wins tool map), with a sensible rationale for not changing library-wide behavior here.

### Issues (all minor — nothing blocking)

**1. The collision error's remediation advice can't work.** The message says "Rename the tools on the server **or load the servers with distinct namePrefix values**" — but this check only ever sees one `fromClient` call, and the prefix is applied uniformly to every tool in it. Prefixing both colliding names with the same string can never separate them (it can only *create* new truncation collisions, e.g. a long prefix pushing distinct suffixes past char 64). For any collision this error reports, the only fix is renaming on the server. Suggest dropping the `namePrefix` clause from the message (it's good advice for cross-server distinctness, where the docs already give it).

**2. `distinctBy` key will silently drift from `ToolDefinition`.** The key enumerates six fields by hand; chimp's `ToolDefinition` (v0.4.0) has exactly those plus `_meta`, so it's complete today — but if chimp adds a field, two genuinely different tools that differ only in that field get silently deduplicated. `definitions.distinctBy(_.copy(_meta = None))` expresses the intent ("identical except `_meta`") and stays correct automatically.

**3. Empty-name guard is bypassed by a prefix.** A tool with an empty original name plus `namePrefix = Some("p")` sanitizes to `"p_"`, passes both checks, and later calls the server with an empty-string name. If empty original names are worth rejecting (the PR argues they are), they should be rejected before prefixing, not only when the *exposed* name ends up empty.

**4. Nit:** the two `new McpToolConversionException(..., null)` calls — a `cause: Throwable = null` default (or an auxiliary constructor) would avoid the literal `null`s.

**5. Worth a release-note line:** sanitization changes LLM-facing names for existing users whose MCP tools have dots/slashes (e.g. `my.tool` → `my_tool`). Those setups were already broken on OpenAI, but on Claude they may have worked, and prompts that reference tools by name will see the new spelling.

### Verdict

Well-scoped, correctly implemented, and unusually well tested — the one thing I'd fix before merge is the misleading `namePrefix` suggestion in the collision error (#1), and #2 is a one-line robustness improvement worth taking. The rest are judgment calls that could ship as-is.

PR #522, opened 2026-07-21
Docs mcp updates

flsh86:


README.md
```diff
@@ -11,7 +11,7 @@ sttp is a family of Scala HTTP-related projects, and currently includes:
 
 * [sttp client](https://github.com/softwaremill/sttp): The Scala HTTP client you always wanted!
 * [sttp tapir](https://github.com/softwaremill/tapir): Typed API descRiptions
-* sttp ai: this project. Non-official Scala client wrapper for OpenAI, Claude (Anthropic), and OpenAI-compatible APIs. Use the power of ChatGPT and Claude inside your code!
+* sttp ai: this project. Non-official Scala client wrapper for OpenAI, Claude (Anthropic), and OpenAI-compatible APIs. Use the power of ChatGPT and Claude inside your code! Agents can also load their tools directly from [Model Context Protocol](https://modelcontextprotocol.io) (MCP) servers.
```
adamw: that's not needed here - too specific

README.md
```diff
@@ -46,7 +46,13 @@ Add the following dependency:
 
 sttp-openai is available for Scala 2.13 and Scala 3
 
-Then head to the [documentation](https://sttp-ai.softwaremill.com) for usage examples: OpenAI and Claude clients, streaming, structured outputs, tool calling, and the agent loop.
+### For loading agent tools from MCP servers
```
adamw: why reference this particular doc page here? I don't get why these changes are here

adamw:
## Review

### What it does

Pure docs/build PR: adds a Claude-backend variant of the MCP example to `docs/agents/mcp.md`, mentions the MCP feature in `README.md` and `CLAUDE.md`, and adds `claude` to the docs project's dependencies so the new mdoc example compiles.

### Verified correct

- **`build.sbt`**: adding `claude` to `docs.dependsOn` is required — the new mdoc block imports `sttp.ai.claude.*` and would fail `compileDocumentation` without it. The right fix, not a workaround.
- **`sbt mcp3/compile` in CLAUDE.md**: verified against sbt directly — `mcp3` resolves and plain `mcp` does *not* (unlike `openai`, where the 2.13 row takes the bare name). The line and its parenthetical explanation are accurate and genuinely needed.
- **The Claude example**: imports, `ClaudeAgent.synchronous(ClaudeConfig.fromEnv, ...)`, and the model id all match the working integration-spec usage; `mdoc:compile-only` will keep it honest.
- The dependency coordinates in the docs (`"com.softwaremill.sttp.ai" %% "mcp"`) match the module's actual artifact name.

### Suggestions

1. **The Claude example is 30 lines to show a 3-line difference.** The intro even says "only the agent construction differs," then repeats the full transport/client/backend/try-finally boilerplate. Consider showing only the differing part (the `ClaudeAgent.synchronous(...)` builder) and stating the rest is identical — or keep the full block if copy-paste runnability is the goal, but then drop the "only the agent construction differs" framing. As written it's the worst of both: long *and* flagged as redundant.
2. **README redundancy**: the new "For loading agent tools from MCP servers" section ends with "see the 'MCP tools' page in the documentation," and the very next paragraph says "head to the documentation for … loading tools from MCP servers." One pointer suffices; consider folding the MCP sentence into the existing "head to the documentation" paragraph and dropping the separate heading — as a subsection of "Add the following dependency" it reads oddly anyway (it contains no dependency coordinates, deferring them to the docs site).
3. **Nit**: `CLAUDE.md` still ends without a trailing newline (pre-existing, but this PR touches that exact line).

Verdict: correct and mergeable as-is; suggestions 1–2 are wording/structure polish.

## Broader review of the MCP-tools docs

### What's good

- **Discoverability is solid.** The page sits logically at the end of the "Agent loop" toctree section (`quickstart → configuration → tools → mcp`), mirroring how a user would progress; README and CLAUDE.md both point to it.
- **`docs/agents/mcp.md` is accurate and honest.** Every claim checked against the merged implementation (#513/#516/#520): pagination, snapshot semantics, name sanitization/truncation rules, the "namePrefix can't fix a same-server collision" caveat, `_meta`-tolerant dedup, null-argument stripping, and both backend caveats all match the code. The known limitation (cross-source collisions silently shadow) is documented rather than hidden.
- The information is user-relevant throughout: observable behavior and remedies, not internals (`rawJsonSchema`, `CustomRaw` correctly stay out of user docs).

### Gaps worth closing

1. **Undecodable schemas aren't mentioned.** `fromClient` also fails with `McpToolConversionException` when a tool's input schema can't be decoded. The docs describe that exception only for name collisions; one sentence under "Results and errors" or "Lifecycle" would complete the failure-mode picture.
2. **No effect-system example.** Both examples use `Identity`/stdio. The text says the HTTP transport "works with any sttp `Backend[F]`", but there's no `fromClient[IO]` sketch, while the agents docs otherwise consistently show the `builder[F]` path.
3. **No runnable MCP example.** `examples/` has agent-loop examples but nothing MCP, and `docs/other/examples.md` only mentions `AgentLoopExample`. Since the feature's pitch is "point it at a real server," a scala-cli-runnable `McpAgentExample` would be the highest-value addition.
4. **`docs/agents/configuration.md`** documents `strictTools` but not *when* you'd need it — the motivation lives only in mcp.md. A cross-reference would connect them.

### Pre-existing typos/staleness in adjacent docs (cheap to fold into this PR)

- `README.md` and `docs/index.md`: "Typed API **descRiptions**" (stray capital R, in both).
- `README.md`: "**sttp-openai** is available for Scala 2.13 and Scala 3" — stale project name, directly adjacent to this PR's new section.
- `docs/agents/tools.md`: `FinishReason` comment says "**Errpr**" (→ `Error`).
- `docs/agents/quickstart.md`: recommends the legacy model `claude-3-haiku-20240307`, while mcp.md and the specs use `claude-haiku-4-5-20251001` — worth aligning.

**Overall**: the MCP docs are in good shape — accurate, appropriately scoped, and findable. Fine to merge after considering the redundancy trims; the improvements that would most help users are the runnable example and the undecodable-schema failure note.

PR #525, opened 2026-07-22
Worktree extra body support

### LINKED ISSUE #477: Add support for nonstandard body params in OpenAI API
Many OpenAI API clients support the `extra_body` parameter that allows you to add additional parameters to the body of the request. See for example: https://reference.langchain.com/python/langchain-openai/chat_models/base/BaseChatOpenAI/extra_body

This is useful for example when calling vLLM where it's crucial to control these parameters.

Could we get support for that in sttp-ai? It's possible to work around this by essentially writing out the body JSON, adding a custom field, and then shoving it back into the API... but that's very cumbersome.

Something like:

```scala
ChatBody(
  messages = ???,
  model = ???,
  extraBody = Map[String, Value](???)
)
```

Note that string support in values is not enough for some APIs, these parameters may be objects or arrays...

flsh86:
## Summary

Adds an `extraBody` escape hatch to `ChatBody`, `CompletionsBody`, and `EmbeddingsBody`, letting callers merge arbitrary JSON parameters into the top level of the serialized request — for parameters that OpenAI-compatible backends (vLLM, etc.) accept but this library has no typed field for (e.g. vLLM's `guided_json`, `top_k`). Mirrors the semantics of the OpenAI Python client's `extra_body`.

- `extraBody: Map[String, Json] = Map.empty` added as the last constructor parameter on all three case classes.
- A single shared helper, `CirceHelpers.mergeExtraBody(fieldName)(json)`, splices the object under `"extra_body"` into the top level of the encoded JSON and drops the wrapper key. Wired into each body's encoder via `.mapJson(mergeExtraBody("extra_body"))`, identically in both the Scala 2 and Scala 3 `OpenAIDerivedCodecs.scala` files.
- On a key collision with a typed field, `extraBody` wins (right-biased `deepMerge`) — documented as a clean override for scalar/array values; object-typed collisions merge recursively rather than replace, which is called out explicitly in the Scaladoc.
- Encode-only; none of the three bodies decode from JSON in this library, so there's no round-trip concern.
- New docs section in `docs/openai/compatible-apis.md` with a vLLM example.

### DIFF
--- a/core/src/main/scala/sttp/ai/core/json/CirceHelpers.scala
+++ b/core/src/main/scala/sttp/ai/core/json/CirceHelpers.scala
@@ -14,4 +14,19 @@ object CirceHelpers {
     j.mapObject(_.filter { case (_, v) => !(v.asArray.exists(_.isEmpty) || v.asObject.exists(_.isEmpty)) })
 
   implicit val omitFalse: Encoder[Option[Boolean]] = Encoder.encodeOption[Boolean].contramap(_.filter(identity))
+
+  /** Splices the JSON object under `fieldName` into the top level of `json`, dropping the `fieldName` wrapper key. Used to implement
+    * `extraBody`-style escape hatches: a request body encodes its `extraBody: Map[String, Json]` field as a nested object under `fieldName`
+    * like any other field, and this then flattens it into siblings of the body's other fields, overriding any of them it collides with
+    * (`deepMerge` is right-biased). A no-op if `fieldName` is absent or not a JSON object.
+    */
+  def mergeExtraBody(fieldName: String)(json: Json): Json =
+    json.asObject match {
+      case Some(obj) =>
+        obj(fieldName).flatMap(_.asObject) match {
+          case Some(extra) => Json.fromJsonObject(obj.remove(fieldName)).deepMerge(Json.fromJsonObject(extra))
+          case None        => json
+        }
+      case None => json
+    }
 }
--- a/core/src/test/scala/sttp/ai/core/json/CirceHelpersSpec.scala
+++ b/core/src/test/scala/sttp/ai/core/json/CirceHelpersSpec.scala
@@ -1,0 +1,48 @@
+package sttp.ai.core.json
+
+import io.circe.Json
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import sttp.ai.core.json.CirceHelpers.mergeExtraBody
+
+class CirceHelpersSpec extends AnyFlatSpec with Matchers {
+
+  "mergeExtraBody" should "splice the object under the given field into the top level, dropping the wrapper key" in {
+    val json = Json.obj(
+      "model" -> Json.fromString("gpt-4o"),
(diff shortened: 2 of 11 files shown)


adamw (commented):
## What the PR does

Adds an `extraBody: Map[String, Json] = Map.empty` escape-hatch field to `ChatBody`, `CompletionsBody`, and `EmbeddingsBody`, letting callers pass arbitrary JSON parameters that get spliced into the top level of the serialized request. This targets OpenAI-compatible backends like vLLM that accept non-standard parameters (`guided_json`, `top_k`), mirroring the OpenAI Python client's `extra_body`. Implementation is one shared helper, `CirceHelpers.mergeExtraBody`, wired into the three encoders via `.mapJson(...)` in both the Scala 2 and Scala 3 codec files, plus unit tests and a docs section with a vLLM example.

## Correctness — solid

- The merge logic is right: the derived encoder emits `extra_body` as a normal nested field, and the helper splices its entries into the top level and drops the wrapper. An empty map serializes as `"extra_body": {}` which the helper removes cleanly, so the default adds nothing to the wire format — confirmed by a test.
- Collision semantics (`base.deepMerge(extra)`, argument-wins in circe) match what's documented: `extraBody` overrides typed fields, with the object-vs-object recursive-merge caveat spelled out honestly in the Scaladoc.
- Scala 2 and Scala 3 codec files are changed identically, which is easy to get wrong here and wasn't.
- Encode-only is correctly claimed — none of the three bodies have decoders, so there's no round-trip asymmetry to worry about.

## Test coverage — good

Four unit tests on the helper (splice, override, absent field, non-object field) plus integration-level tests through the real encoders for all three bodies, including verifying the `extra_body` key never leaks to the wire. No gaps I'd block on.

## Issues and nits (none blocking)

1. **Explicit `null` values are silently dropped.** `OpenAIJson.asJson` runs `deepDropNullValues` after encoding (`OpenAIJson.scala:47`), so `extraBody = Map("x" -> Json.Null)` never reaches the wire. If someone uses `extra_body` to explicitly null-out a parameter (as the Python client allows), it won't work. Worth one sentence in the Scaladoc.
2. **Map keys bypass snake_case conversion.** Typed fields get `extraBody` → `extra_body` automatically, but keys inside the map are sent verbatim — a caller writing `"topK"` gets no help. The docs example uses correct wire names, but the Scaladoc should state "keys are used as-is".
3. **Binary compatibility:** appending a case-class field breaks bincompat (constructor/`copy`/`unapply`). No MiMa in this build and the library is 0.x, so presumably accepted practice — just flagging it for the release notes.
4. **PR title** looks like a leaked worktree branch name ("Worktree extra body support") — consider renaming to something like "Add extraBody support for OpenAI-compatible request parameters" before merge, since it becomes the squash-commit message.
5. **Scope choice:** `ModerationsBody` and `SpeechRequestBody` don't get the field. Reasonable for a first pass, and the pattern is now trivially extensible if requested.

## Verdict

Well-scoped, cleanly implemented, honestly documented, and properly tested on both Scala versions. The only pre-merge ask is the title fix; items 1–2 are optional doc improvements.

PR #529, opened 2026-07-23
No tools final iteration

### LINKED ISSUE #480: Don't allow the agent to call tools on the last iteration
The system prompt for the agent specifies that the agent should provide the final answer on the last iteration "even if the result is partial, approximate, or only a summary" and yet the request still specifies tools to use. If we expect no tool calls, we should not provide any tools to use in the last call.

flsh86:
## Problem

The agent loop offered tools on every model call, including the last allowed iteration (iteration == maxIterations - 1). The default system prompt (rule 4) already asked the model to give a final answer on that last iteration, but nothing enforced it: if the model made a tool call there, the results were computed and then discarded — the loop hits the iteration cap before the model ever sees them — and the "final answer" degraded to a raw tool-result string.

## Change

On the last allowed iteration, the request is now sent without tools, forcing the model to return a text answer.

- AgentBackend.sendRequest gains an explicit third parameter includeTools: Boolean — the loop is the only place that knows the iteration count, so it owns the decision.
- OpenAIAgentBackend / ClaudeAgentBackend attach tools only when includeTools && convertedTools.nonEmpty.
- Agent.loop computes isLastIteration and passes includeTools = !isLastIteration. On the last iteration it terminates with FinishReason.MaxIterations, using the model's text answer (or falling back to extractFinalAnswer(history) if that text is empty), and ignores any spurious tool calls in that response (they are neither executed nor recorded). The top-of-loop iteration >= maxIterations guard is retained purely as a safety net for maxIterations <= 0.

FinishReason.MaxIterations now means "the answer was forced on the final iteration" — honest telemetry that the run was truncated rather than the model finishing naturally.

## ⚠️ Breaking change

AgentBackend.sendRequest gains a required includeTools: Boolean parameter. This is source- and binary-breaking for anyone who has implemented AgentBackend for a custom provider — a trivial one-parameter addition. All in-repo implementers (OpenAI, Claude, test stub) are updated in this PR.

Behavior note

maxIterations now reserves the last iteration for the final answer, so tools are available for the first maxIterations - 1 iterations. With maxIterations = 1 the agent never gets to use its tools. Documented in docs/agents/configuration.md.

## Tests & verification

- AgentSpec (loop, stub backend): asserts the last iteration is called with includeTools = false (Seq(true, true, false)), its tool calls are not executed, and the forced text answer is used. Two existing tests updated to the corrected behavior.
- Backend specs (OpenAI + Claude): capture the serialized request body and assert tools is present when includeTools = true, absent when false.
- Integration test (shared base, runs for OpenAI/Claude/MCP): with maxIterations = 2 the model calls a tool on iteration 1 and is forced to answer tool-free on iteration 2 — validating the APIs accept a tools-omitted request whose history contains tool_use/tool_result entries. Auto-skips without keys.
- All JVM suites pass on Scala 2.13 and 3.3.
- Verified live against the real OpenAI API: the integration test passes (1/1), and the added examples/FinalIterationNoToolsExample logs tools offered: true → true → false and completes with MaxIterations and no API error — confirming OpenAI accepts a tools-omitted request carrying prior tool history. (Claude not yet exercised live — same code path, per-provider risk.)

### DIFF
--- a/claude/src/main/scala/sttp/ai/claude/agent/ClaudeAgent.scala
+++ b/claude/src/main/scala/sttp/ai/claude/agent/ClaudeAgent.scala
@@ -81,7 +81,8 @@ private[claude] class ClaudeAgentBackend[F[_]](
 
   override def sendRequest(
       history: ConversationHistory,
-      backend: Backend[F]
+      backend: Backend[F],
+      includeTools: Boolean
   ): F[AgentResponse] = {
     val messages = buildMessages(history)
     val request = MessageRequest(
@@ -89,7 +90,7 @@ private[claude] class ClaudeAgentBackend[F[_]](
       messages = messages.toList,
       maxTokens = 4096,
       system = systemPrompt,
-      tools = if (convertedTools.nonEmpty) Some(convertedTools.toList) else None,
+      tools = if (includeTools && convertedTools.nonEmpty) Some(convertedTools.toList) else None,
       outputConfig = outputConfig
     )
 
--- a/claude/src/test/scala/sttp/ai/claude/agent/ClaudeAgentBackendSpec.scala
+++ b/claude/src/test/scala/sttp/ai/claude/agent/ClaudeAgentBackendSpec.scala
@@ -126,7 +126,7 @@ class ClaudeAgentBackendSpec extends AnyFlatSpec with Matchers with EitherValues
       captured.set(request)
       ResponseStub.adjust(minimalMessageResponse, StatusCode.Ok)
     }
-    backend.sendRequest(ConversationHistory.withInitialPrompt("hello"), httpStub, includeTools = includeTools)
+    backend.sendRequest(ConversationHistory.withInitialPrompt("hello"), httpStub, includeTools = includeTools): Unit
     captured.get().body match {
       case StringBody(s, _, _) => s
       case other               => fail(s"expected StringBody, got $other")
--- a/claude/src/test/scala/sttp/ai/claude/agent/ClaudeAgentBackendSpec.scala
+++ b/claude/src/test/scala/sttp/ai/claude/agent/ClaudeAgentBackendSpec.scala
@@ -9,9 +9,15 @@ import sttp.ai.claude.ClaudeClient
 import sttp.ai.claude.config.ClaudeConfig
 import sttp.ai.claude.models.Tool
 import sttp.ai.core.agent.AgentTool
+import sttp.ai.core.agent.ConversationHistory
 import sttp.apispec.Schema
+import sttp.client4._
+import sttp.client4.testing.ResponseStub
+import sttp.model.StatusCode
 import sttp.monad.IdentityMonad
 import sttp.shared.Identity
+
+import java.util.concurrent.atomic.AtomicReference
 
 class ClaudeAgentBackendSpec extends AnyFlatSpec with Matchers with EitherValues {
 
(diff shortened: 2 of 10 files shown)


adamw:
# Code Review

## Overview

The PR fixes a real defect in the shared agent loop: previously, tools were offered on every model call including the last allowed iteration, so if the model made a tool call there, the results were computed, recorded, and then discarded — the loop hit the iteration cap and the "final answer" degraded to a raw tool-result string via `extractFinalAnswer`. The change withholds tools on the last iteration (`AgentBackend.sendRequest` gains an `includeTools: Boolean` parameter; `Agent.loop` passes `!isLastIteration`), forcing a text answer, and terminates that iteration with `FinishReason.MaxIterations`.

The change is small, well-scoped, well-tested at the unit level, and the loop logic in `Agent.scala` is correct — I traced the `MaxTokens`, last-iteration, natural-stop, and `maxIterations <= 0` paths and the new branch ordering is right, including the fallback to `extractFinalAnswer(history)` when the forced response has empty text. The updated test expectations (`toolCalls` size 4 instead of 5, forced text answer instead of discarded tool result) match the new semantics precisely.

## Main concern: the Claude path is unverified against a known API constraint

On the last iteration, `ClaudeAgentBackend` now sends a request with **no `tools` parameter while the message history still contains `tool_use` and `tool_result` content blocks**. The Anthropic API has historically rejected exactly this shape with a 400: *"Requests which include `tool_use` or `tool_result` blocks must define tools."* The current tool-use documentation neither confirms nor denies that this constraint still holds — and the PR description explicitly states Claude was **not exercised live** ("same code path, per-provider risk"). The OpenAI path was verified live and is fine.

If the constraint still applies, the failure mode is the worst possible one: the run throws an API error at exactly the moment it's supposed to deliver the final answer, and only on runs that actually exhaust their iterations — easy to miss in casual testing. **I'd block merge on running `ClaudeAgentIntegrationSpec` (the new "force a tool-free final answer" test) against the real API once.** That single test run resolves the question.

**Safer alternative worth considering:** instead of omitting `tools`, keep them and set tool choice to "none":
- OpenAI: `ChatBody.toolChoice = Some(ToolChoice.None)` — already supported (`openai/.../chat/message/ToolChoice.scala`).
- Anthropic: `tool_choice: {"type": "none"}` — supported by the API, though this repo's `MessageRequest` doesn't model `tool_choice` yet, so it would need a small addition.

This sidesteps the tools-with-history question entirely, and has a second benefit:

## Prompt-cache cost on the final request

Anthropic's prompt cache is a prefix match with tools rendered first — **removing the tool definitions invalidates the entire cached prefix** (tools + system + full conversation). The final iteration carries the longest history of the whole run, so with this design it is always billed fully uncached. `tool_choice: none` preserves the tools/system cache tiers. Not a correctness issue, but for long agent runs it's a measurable cost regression on the most expensive request. (The library doesn't set `cache_control` today, so this is forward-looking, but the request-shape decision being made here is what determines it.)

## Semantics worth noting (intentional, but with an edge case)

On the last iteration, the `isLastIteration` branch is checked before the `toolCalls.isEmpty` branch, so `FinishReason.NaturalStop` is unreachable there — even a model that would have finished naturally reports `MaxIterations`. Since tools were withheld, the two genuinely can't be distinguished, so this is defensible. The degenerate case: with `maxIterations = 1`, every run — including trivial no-tool Q&A — now reports `MaxIterations`, and tools are never offered at all. The docs change covers the tools part clearly; consider whether `AgentBuilder` should log/warn (or the docs should say more loudly) when `maxIterations == 1` is combined with a non-empty tool set, since that configuration silently makes all tools dead weight.

## Minor points

- **Example's scala-cli directive is misleading** (`examples/FinalIterationNoToolsExample.scala:1`): `//> using dep com.softwaremill.sttp.ai::openai:0.5.2` pins the published release, which does *not* contain this change — anyone running it with scala-cli (the documented way to run examples) would observe `tools offered: true, true, true`. The scaladoc does say to run via `sbt examples/runMain`, but the directive contradicts it. Either point it at the SNAPSHOT + `//> using repository ivy2Local`, or bump it at release time.
- **Source compatibility**: `includeTools` could have been given a default value (`includeTools: Boolean = true`) to keep external `AgentBackend` *callers* source-compatible; implementers break either way, and the explicit parameter forces implementers to make a conscious choice, so the PR's approach is defensible — just worth an explicit callout in the release notes.
- **Comment quality**: the safety-net comment above the `iteration >= config.maxIterations` guard in `Agent.scala` is genuinely useful (it explains why apparently-dead code is kept). Good.
- **Tests**: the backend specs capturing the serialized request body and asserting on `"tools"` presence/absence are the right level of verification for the serialization concern, and the shared integration test exercising tools-omitted-with-tool-history is exactly the scenario that matters. Assertions like `should not include "\"tools\""` are string-fragile but acceptable here.

## Verdict

Approve with one condition: **run the Claude integration test live before merging** (or switch the mechanism to `tool_choice: none`, which also fixes the cache-invalidation cost). Everything else — loop logic, test updates, docs, the example — is solid and the behavioral change itself is a clear improvement over silently discarding the last iteration's tool results.

PR #534, opened 2026-07-27
Handle MaxIterations/TokenLimit gracefully in Agent.runAs

### LINKED ISSUE #481: Agent loop with structured output: handle MaxIterations/TokenLimit errors gracefully
As it stands, if you use `runAs[T]` and the loop finishes with `FinishReason.MaxIterations` or `FinishReason.TokenLimit`, the `finalAnswer`, which has a type of `Either[AgentParseError, T]`, will result in an `AgentParseError`. It makes no sense to attempt parsing in this case so a more graceful handling would be preferred.

flsh86:
## Problem

`runAs[T]` always attempted `decode[T]` on the final answer, regardless of how the loop finished. When the loop ended with `FinishReason.TokenLimit` (truncated output) or `FinishReason.MaxIterations`, a decode failure surfaced as `AgentParseError` — indistinguishable from "the model returned malformed JSON on a clean stop". For `TokenLimit` the parse attempt is meaningless: the output is truncated by construction.

## Changes

- `finalAnswer`'s `Left` type widens from `AgentParseError` to a new sealed trait `AgentFailure` with two cases:
  - `AgentParseError(rawAnswer, cause)` — unchanged name and fields; a clean stop whose answer couldn't be parsed as `T`.
  - `AgentIncomplete(rawAnswer, finishReason, parseError: Option[Throwable])` — the loop was cut short.
- Decision table in `runAs[T]`:

  | `finishReason` | Behavior |
  |---|---|
  | `NaturalStop` | Parse; failure → `AgentParseError` (unchanged) |
  | `MaxIterations` | Parse (the forced final iteration is schema-guided and tool-free, so often valid); failure → `AgentIncomplete` with `parseError = Some(cause)` |
  | `TokenLimit` / `Error` | Never parse → `AgentIncomplete` with `parseError = None` |

- Docs (`docs/agents/tools.md`) updated to describe both failure cases. Because `MaxIterations` still parses, `Right(t)` can coexist with `finishReason = MaxIterations` — documented, with a pointer to `AgentResult.finishReason` for callers that need to distinguish capped runs.
- `TypedAgentLoopExample` keeps a release-compatible `case Left(err)` match (CI compiles examples with scala-cli against the published artifact, which doesn't contain `AgentIncomplete` yet); the intended two-case match is included as a TODO to swap in after the next release.

## ⚠️ Breaking change

Source-breaking for `runAs[T]` callers: the `Left` type widens from `AgentParseError` to `AgentFailure`, which only exposes `rawAnswer`. Code like `case Left(err) => err.cause` no longer compiles.

**Migration:** match on the cases — `case Left(AgentParseError(raw, cause)) => ...` / `case Left(AgentIncomplete(raw, finishReason, parseError)) => ...`. Code that only uses `.isRight`/`.toOption`/`rawAnswer`, or that constructs/matches `AgentParseError` directly, compiles unchanged.

## Testing

- New/updated `AgentSpec` unit tests cover every row of the table, including that `TokenLimit` skips the parse even when the raw text happens to be valid JSON.
- `core/test` and `core3/test`: 47/47 pass; `compileDocumentation` passes; examples verified with scala-cli against the published 0.5.4; scalafmt clean.

### DIFF
--- a/core/src/main/scala/sttp/ai/core/agent/Agent.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/Agent.scala
@@ -93,10 +93,17 @@ class Agent[F[_]](
 
   def runAs[T](
       initialPrompt: String
-  )(backend: Backend[F])(implicit r: Decoder[T]): F[AgentResult[Either[AgentParseError, T]]] =
+  )(backend: Backend[F])(implicit r: Decoder[T]): F[AgentResult[Either[AgentFailure, T]]] =
     monad.map(run(initialPrompt)(backend)) { res =>
-      val parsed: Either[AgentParseError, T] =
-        decode[T](res.finalAnswer).left.map(e => AgentParseError(res.finalAnswer, e))
+      val parsed: Either[AgentFailure, T] = res.finishReason match {
+        case FinishReason.NaturalStop =>
+          decode[T](res.finalAnswer).left.map(e => AgentParseError(res.finalAnswer, e))
+        case FinishReason.MaxIterations =>
+          // The last iteration forces a no-tools, schema-guided answer, so it may still be valid.
+          decode[T](res.finalAnswer).left.map(e => AgentIncomplete(res.finalAnswer, res.finishReason, Some(e)))
+        case FinishReason.TokenLimit | FinishReason.Error(_) =>
+          Left(AgentIncomplete(res.finalAnswer, res.finishReason, parseError = None))
+      }
       AgentResult(parsed, res.iterations, res.toolCalls, res.finishReason)
     }
 
--- a/core/src/main/scala/sttp/ai/core/agent/AgentResult.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/AgentResult.scala
@@ -24,7 +24,17 @@ final case class AgentResult[T](
     finishReason: FinishReason
 )
 
+sealed trait AgentFailure {
+  def rawAnswer: String
+}
+
 final case class AgentParseError(
     rawAnswer: String,
     cause: Throwable
-)
+) extends AgentFailure
+
+final case class AgentIncomplete(
+    rawAnswer: String,
+    finishReason: FinishReason,
+    parseError: Option[Throwable]
+) extends AgentFailure
--- a/core/src/test/scala/sttp/ai/core/agent/AgentSpec.scala
+++ b/core/src/test/scala/sttp/ai/core/agent/AgentSpec.scala
@@ -437,12 +437,34 @@ class AgentSpec extends AnyFlatSpec with Matchers with OptionValues {
     result.iterations shouldBe 1: Unit
     result.finishReason shouldBe FinishReason.NaturalStop: Unit
     result.finalAnswer.isLeft shouldBe true: Unit
-    val err = result.finalAnswer.left.toOption.get
-    err.rawAnswer should include("wrong"): Unit
-    err.cause should not be null
+    result.finalAnswer.left.toOption.get match {
+      case AgentParseError(rawAnswer, cause) =>
+        rawAnswer should include("wrong"): Unit
+        cause should not be null
+      case other => fail(s"Expected AgentParseError, got $other")
+    }
   }
 
-  it should "return Left(AgentParseError) on the maxIterations path where the capped answer is not schema-shaped" in {
+  it should "return Right(T) on the maxIterations path when the forced final answer is schema-shaped" in {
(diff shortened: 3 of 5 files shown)


adamw:
# Code Review

## Overview

Previously `runAs[T]` unconditionally ran `decode[T]` on the final answer, so a truncated (`TokenLimit`) or capped (`MaxIterations`) run surfaced as `AgentParseError` — indistinguishable from "the model produced malformed JSON on a clean stop". The PR widens the `Left` type to a sealed `AgentFailure` with two cases: `AgentParseError` (unchanged, clean stop + parse failure) and `AgentIncomplete` (loop cut short, carrying the `finishReason` and an optional parse error). `MaxIterations` still attempts the parse (sensible, since #529 made the last iteration a forced schema-guided, tool-free answer); `TokenLimit`/`Error` never parse.

## Correctness

- The new match in `Agent.runAs` is **exhaustive** against the actual `FinishReason` ADT (`MaxIterations`, `NaturalStop`, `TokenLimit`, `Error`). Being a sealed trait, a future new case would at least warn at compile time — the right safety net.
- `FinishReason.Error` is **defined but never produced anywhere in the loop** — the `TokenLimit | Error(_)` arm's `Error` half is currently dead code, kept only for exhaustiveness. Fine, but worth knowing it's untestable through the public loop today.
- The decision table itself is sound, and treating a parse success under `MaxIterations` as `Right(t)` (rather than forcing a failure) is the right call — the docs note explaining that `Right` can coexist with a capped `finishReason` addresses the one confusion this could cause.

## Design observations (minor, non-blocking)

- **`TokenLimit` discards a possibly-valid answer in one rare case**: if `max_tokens` happens to be hit exactly at the end of a complete JSON document, the output parses fine but the PR still returns `AgentIncomplete` (the new test explicitly locks this in). Truncated JSON is syntactically invalid in virtually all other positions, so the discarded case is genuinely rare, and the conservative "truncated by construction" stance is defensible — but an alternative (attempt the parse, mirroring the `MaxIterations` row) would rescue that case at essentially no risk of accepting semantically incomplete data, since mid-document truncation can't parse. Worth a conscious decision rather than an accident; the PR body suggests it was conscious.
- **`AgentIncomplete.finishReason` duplicates `AgentResult.finishReason`.** It makes the failure value self-describing in a bare `Left(...)` match, which is a real ergonomic win, but the constructor doesn't prevent nonsensical values (`AgentIncomplete(_, NaturalStop, _)` compiles). Not worth restricting with a subtype — just noting the type doesn't enforce the invariant.

## Compatibility

Adding `extends AgentFailure` to `AgentParseError` is binary-compatible; the source break is confined to `runAs` callers that accessed `err.cause` on the widened type. The migration note in the PR body is accurate — code using only `rawAnswer` (which the trait exposes) compiles unchanged. Appropriate for a 0.x library, and clearly flagged.

## Tests & docs

- Coverage maps one-to-one onto the decision table: NaturalStop parse failure (updated), MaxIterations parse success (new), MaxIterations parse failure now yielding `AgentIncomplete` with `parseError` defined (updated), and TokenLimit skipping the parse even for valid JSON (new). That last test is the most valuable — it pins the deliberate behavior rather than letting it look like an oversight.
- The docs update (`docs/agents/tools.md`) is clear, uses the new two-case match (compiled by mdoc against the local build, so it's verified), and documents the `Right`-with-`MaxIterations` subtlety.

## One smell: the commented-out code in `TypedAgentLoopExample`

The example keeps a release-compatible single-case match with the intended two-case match as a commented TODO, because CI compiles examples with scala-cli against the published artifact. The retained `case Left(err) => ... ${err.rawAnswer}` is a neat trick — it compiles against both the old `AgentParseError` and the new `AgentFailure` trait since both expose `rawAnswer`. But commented-out code tends to rot silently after the release that would unblock it. Suggest tracking the swap in an issue (or a release checklist) instead of relying on someone rediscovering the comment — this is the same published-artifact-lag problem flagged on #529's example, so it may be worth a one-time decision about how examples handle unreleased API (e.g., `//> using repository ivy2Local` + SNAPSHOT in CI).

## Verdict

**Approve.** Correct, exhaustively matched against the real ADT, every row of the behavior table is pinned by a test, the breaking change is honestly documented with a migration path, and the design choices (parse on `MaxIterations`, skip on `TokenLimit`) are reasonable. The only follow-ups are process-level: track the example's deferred match swap, and optionally reconsider the exact-boundary `TokenLimit` case — neither blocks merge.

PR #537, opened 2026-07-27
Gemini/01 core

flsh86:
First PR of a 6-PR stack adding a native Google Gemini client module targeting the [Interactions API](https://ai.google.dev/gemini-api/docs/interactions-overview) — Google's recommended surface for new projects (`generateContent` is now legacy). The module mirrors the existing `claude` module's architecture.

Until now Gemini was only reachable through the OpenAI-compatible endpoint, which exposes just the chat-completions surface. This stack gives it first-class treatment: dedicated client, sync wrapper, agent-loop backend, streaming for all five effect systems, and docs.

## This PR: module skeleton + full data model (no HTTP client yet)

**Build wiring**
- `gemini` projectMatrix cloned from `claude`'s: JVM rows for Scala 3 + 2.13, Native row for Scala 3, `dependsOn(core % "compile->compile;test->test")`, added to aggregates.

**Config**
- `GeminiConfig` extending core's `AIClientConfig`: `x-goog-api-key` auth header, `GEMINI_API_KEY` / `GEMINI_BASE_URL` env vars, default base URL `https://generativelanguage.googleapis.com`.

**Models** (package `sttp.ai.gemini.models`)
- `Content` (text/image/audio/video/document), `Step` (`user_input` / `model_output` / `function_call` / `function_result` / `thought`, plus an `Unknown` fallback so new server-side step types can never break deserialization)
- `InteractionInput` — the request `input` union: plain string or a list of steps (stateless conversation replay)
- `Tool` — `Function` with a **verbatim raw JSON-schema `parameters`** (so MCP-sourced schemas survive byte-faithful, incl. legitimate `null`s), plus `GoogleSearch` / `CodeExecution` built-ins
- `ResponseFormat` — `Text` or `JsonSchema(schema)` for structured output
- `GenerationConfig`, `SafetySetting`, `Usage`, `InteractionStatus` (open enum: unknown statuses decode as `Other`), `GeminiModel`
- `InteractionRequest` / `InteractionResponse` / `ErrorResponse` / `InteractionStreamEvent` (open SSE envelope — unknown `event_type`s decode fine)

**JSON**
- circe with core's shared `CirceConfiguration` (snake_case + `type` discriminator); derived codecs per Scala version (`io.circe.derivation` on 3, `circe-generic-extras` on 2.13), hand-written codecs where derivation can't express the wire shape (`Tool`, `ResponseFormat`, `InteractionInput`, `InteractionStatus`, `ErrorDetail`).

**Tests:** wire-format round-trips asserted against literal JSON (not just structural equality), unknown-status/step/event tolerance, verbatim tool-schema preservation.

## Validated against the live API

The whole stack was exercised end-to-end with a real `GEMINI_API_KEY` (10/10 integration tests in PR 5 pass). Several contract facts in this PR come from live responses rather than the docs, which turned out to be wrong or incomplete:
- `response_format` takes the JSON schema **directly** (the documented `{"type":"json_schema", ...}` wrapper is rejected)
- `store=false` responses carry **no `id`** → `InteractionResponse.id: Option[String]`
- responses contain a `thought` step type; error bodies can carry `code` as a string (`"not_found"`) or a number — the decoder accepts both

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -35,6 +35,7 @@ lazy val root = (project in file("."))
 lazy val allAgregates = core.projectRefs ++
   openai.projectRefs ++
   claude.projectRefs ++
+  gemini.projectRefs ++
   fs2.projectRefs ++
   zio.projectRefs ++
   pekko.projectRefs ++
@@ -81,6 +82,23 @@ lazy val openai = (projectMatrix in file("openai"))
   .dependsOn(core % "compile->compile;test->test")
 
 lazy val claude = (projectMatrix in file("claude"))
+  .jvmPlatform(
+    scalaVersions = scala3 ++ scala2 // Scala 3 first priority
+  )
+  .nativePlatform(
+    scalaVersions = scala3
+  )
+  .settings(commonSettings: _*)
+  .settings(
+    libraryDependencies ++=
+      Seq(Libraries.tapirApispecDocs.value) ++
+        Libraries.sttpApispec.value ++ Libraries.circe.value ++
+        Libraries.sttpClient.value ++ Seq(Libraries.scalaTest.value),
+    libraryDependencies ++= (if (scalaVersion.value.startsWith("2.")) Seq(Libraries.circeGenericExtras.value) else Seq.empty)
+  )
+  .dependsOn(core % "compile->compile;test->test")
+
+lazy val gemini = (projectMatrix in file("gemini"))
   .jvmPlatform(
     scalaVersions = scala3 ++ scala2 // Scala 3 first priority
   )
--- a/gemini/src/main/scala-2/sttp/ai/gemini/json/GeminiDerivedCodecs.scala
+++ b/gemini/src/main/scala-2/sttp/ai/gemini/json/GeminiDerivedCodecs.scala
@@ -4,6 +4,7 @@ import io.circe.Codec
 import io.circe.generic.extras.semiauto.deriveConfiguredCodec
 import sttp.ai.gemini.models._
 import sttp.ai.gemini.requests.InteractionRequest
+import sttp.ai.gemini.responses._
 import sttp.ai.core.json.CirceConfiguration.jsonConfiguration
 import GeminiManualCodecs._
 
@@ -15,4 +16,11 @@ object GeminiDerivedCodecs {
   implicit val safetySettingCodec: Codec[SafetySetting] = deriveConfiguredCodec
   implicit val stepCodec: Codec[Step] = deriveConfiguredCodec
   implicit val interactionRequestCodec: Codec[InteractionRequest] = deriveConfiguredCodec
+
+  implicit val interactionResponseCodec: Codec[InteractionResponse] = deriveConfiguredCodec
+  implicit val errorDetailCodec: Codec[ErrorDetail] = deriveConfiguredCodec
+  implicit val errorResponseCodec: Codec[ErrorResponse] = deriveConfiguredCodec
+  implicit val streamErrorCodec: Codec[StreamError] = deriveConfiguredCodec
+  implicit val streamMetadataCodec: Codec[StreamMetadata] = deriveConfiguredCodec
+  implicit val interactionStreamEventCodec: Codec[InteractionStreamEvent] = deriveConfiguredCodec
 }
--- a/gemini/src/main/scala-2/sttp/ai/gemini/json/GeminiDerivedCodecs.scala
+++ b/gemini/src/main/scala-2/sttp/ai/gemini/json/GeminiDerivedCodecs.scala
@@ -3,7 +3,9 @@ package sttp.ai.gemini.json
 import io.circe.Codec
 import io.circe.generic.extras.semiauto.deriveConfiguredCodec
 import sttp.ai.gemini.models._
+import sttp.ai.gemini.requests.InteractionRequest
 import sttp.ai.core.json.CirceConfiguration.jsonConfiguration
+import GeminiManualCodecs._
 
 object GeminiDerivedCodecs {
 
@@ -11,4 +13,6 @@ object GeminiDerivedCodecs {
   implicit val contentCodec: Codec[Content] = deriveConfiguredCodec
   implicit val generationConfigCodec: Codec[GenerationConfig] = deriveConfiguredCodec
   implicit val safetySettingCodec: Codec[SafetySetting] = deriveConfiguredCodec
+  implicit val stepCodec: Codec[Step] = deriveConfiguredCodec
+  implicit val interactionRequestCodec: Codec[InteractionRequest] = deriveConfiguredCodec
 }
--- a/gemini/src/main/scala-2/sttp/ai/gemini/json/GeminiDerivedCodecs.scala
+++ b/gemini/src/main/scala-2/sttp/ai/gemini/json/GeminiDerivedCodecs.scala
@@ -1,6 +1,6 @@
 package sttp.ai.gemini.json
 
-import io.circe.Codec
+import io.circe.{Codec, Decoder, Encoder, Json}
 import io.circe.generic.extras.semiauto.deriveConfiguredCodec
 import sttp.ai.gemini.models._
 import sttp.ai.gemini.requests.InteractionRequest
@@ -14,11 +14,21 @@ object GeminiDerivedCodecs {
   implicit val contentCodec: Codec[Content] = deriveConfiguredCodec
   implicit val generationConfigCodec: Codec[GenerationConfig] = deriveConfiguredCodec
   implicit val safetySettingCodec: Codec[SafetySetting] = deriveConfiguredCodec
-  implicit val stepCodec: Codec[Step] = deriveConfiguredCodec
+
+  // Derived codec for the known Step discriminators only; unrecognized `type` values fall back to Step.Unknown so that
+  // decoding a response never fails just because the API introduced a new step type.
+  private val stepBaseCodec: Codec[Step] = deriveConfiguredCodec
+  implicit val stepCodec: Codec[Step] = Codec.from(
+    stepBaseCodec.or(Decoder[Json].map(Step.Unknown.apply)),
+    Encoder.instance {
+      case Step.Unknown(raw) => raw
+      case other             => stepBaseCodec(other)
+    }
+  )
+
   implicit val interactionRequestCodec: Codec[InteractionRequest] = deriveConfiguredCodec
 
   implicit val interactionResponseCodec: Codec[InteractionResponse] = deriveConfiguredCodec
-  implicit val errorDetailCodec: Codec[ErrorDetail] = deriveConfiguredCodec
   implicit val errorResponseCodec: Codec[ErrorResponse] = deriveConfiguredCodec
   implicit val streamErrorCodec: Codec[StreamError] = deriveConfiguredCodec
   implicit val streamMetadataCodec: Codec[StreamMetadata] = deriveConfiguredCodec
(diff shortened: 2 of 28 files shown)


adamw:
## Review — PR 1/6 (core module: build wiring, config, data model, codecs)

Reviewed the whole 6-PR stack. Overall this is high-quality work: the module mirrors `claude` faithfully, package-by-functionality is respected, the dependency direction is clean (`gemini` → `core` only), the PR split is sensible and no PR forward-references a later one. I verified the tip of the stack builds and is green: `scalafmtCheckAll` + `Test/compile` + `gemini/test` → 50 passed, 10 cancelled (integration, no key).

Findings for this PR first, then a section of **global remarks** that apply across the stack (posted here since this is the base of the stack).

---

### 1. `Step` fallback decoder swallows *all* decoding failures, not just unknown step types — high

`gemini/src/main/scala-3/sttp/ai/gemini/json/GeminiDerivedCodecs.scala:21-22` (identical in `scala-2/`):

```scala
stepBaseCodec.or(Decoder[Json].map(Step.Unknown.apply))
```

`Decoder.or` falls back on **any** `DecodingFailure`, and `Decoder[Json]` never fails. So a *known* step that fails on a nested field silently becomes `Step.Unknown` instead of surfacing an error. The comment above it says the fallback is for "unrecognized `type` values" — it does much more than that.

This matters because `Content` (`models/Content.scala:4-31`) is a **closed** ADT (Text/Image/Audio/Video/Document) with no fallback case, while the client explicitly offers `Tool.GoogleSearch` and `Tool.CodeExecution` — exactly the features whose responses carry non-text content blocks.

I confirmed both consequences by running against the built module:

```
{"status":"completed","steps":[{"type":"model_output","content":[
   {"type":"text","text":"hi"},{"type":"executable_code","code":"x=1"}]}]}
→ InteractionResponse(..., List(Unknown({...})))
→ outputText == ""            // the text was right there in the payload

{"status":"requires_action","steps":[{"type":"function_call","name":"get_weather","arguments":{}}]}   // no "id"
→ List(Unknown({...}))
→ functionCalls == List()     // agent loop reads this as end-of-turn
```

Downstream: `GeminiSyncClient.createInteractionAs` throws "Failed to parse structured output" on an empty string, and `GeminiAgentBackend` (PR 3) returns `AgentResponse("", Nil, StopReason.EndTurn)` — a successful-looking empty answer instead of a loud failure.

Suggested fix — gate the fallback on the discriminator so real field-level failures propagate, and give `Content` the same forward-compatibility the steps were meant to have:

```scala
private val KnownStepTypes = Set("user_input", "model_output", "function_call", "function_result", "thought")

implicit val stepCodec: Codec[Step] = Codec.from(
  Decoder.instance { c =>
    c.get[Option[String]]("type").flatMap {
      case Some(t) if KnownStepTypes.contains(t) => stepBaseCodec(c)   // real failures propagate
      case _                                     => Right(Step.Unknown(c.value))
    }
  },
  Encoder.instance {
    case Step.Unknown(raw) => raw
    case other             => stepBaseCodec(other)
  }
)
```

plus a `case class Unknown(raw: Json) extends Content`. Note this has to change in **both** version files — see #4.

### 2. `ResponseFormat` decoder rejects schemas whose top-level `type` is not a string — medium

`json/GeminiManualCodecs.scala:62`: `c.get[Option[String]]("type")` returns `Left(DecodingFailure)` — not `None` — for a legal JSON Schema like `{"type":["object","null"]}`, so `ResponseFormat.JsonSchema` fails to round-trip. Encode-only paths are unaffected (responses never carry `response_format`), so nothing live breaks, but the round-trip tests in `ResponseFormatSpec` would not catch it either.

```scala
c.get[Option[Json]]("type").map(_.flatMap(_.asString)).flatMap { ... }
```

### 3. `GeminiManualCodecs` ↔ `GeminiDerivedCodecs` recreate the class-init cycle from 55da3fc — medium

`GeminiDerivedCodecs.scala:9` imports `GeminiManualCodecs._`, and `GeminiManualCodecs.scala:77,82` reference `GeminiDerivedCodecs.stepCodec`. That is exactly the mutual-object shape behind commit `55da3fc` ("Fix circular class-init deadlock between OpenAIManualCodecs and OpenAIDerivedCodecs").

The current code **is** safe — both back-references sit inside `Decoder.instance`/`Encoder.instance` bodies, so they are deferred past `<clinit>`. But `OpenAIManualCodecs.scala:34-40` carries an explicit doc comment stating that one-way-init invariant and this file carries none; a future refactor hoisting `Decoder.decodeList(GeminiDerivedCodecs.stepCodec)` to the val's RHS would silently reintroduce a deadlock under parallel suites.

Preferred fix: move `interactionInputCodec` into `GeminiDerivedCodecs` (next to `stepCodec`, the only thing it needs) and delete the back-reference — that matches `claude`, which has no cycle at all. Fallback: copy the OpenAI invariant comment.

### 4. The two `GeminiDerivedCodecs.scala` files duplicate the non-trivial part — low/medium

They differ only in the import and `deriveConfiguredCodec` vs `ConfiguredCodec.derived`; the ~8-line `stepCodec` fallback wrapper and its comment are copy-pasted with nothing enforcing lockstep — which is why #1 has to be fixed twice. Consider keeping only `private[json] val stepBaseCodec` + the plain derived vals in the version dirs, and building the `Unknown` wrapper once in a shared object under `src/main/scala`.

(Verified clean, FWIW: I had the two derivations compared side-by-side under `core.json.CirceConfiguration` with the gemini model shapes — encoder output is byte-identical across 2.13/3 for every `Step` constructor, nested `Content`, and snake_cased `Usage`, and the decoders agree on defaults/nulls/extra fields. No wire-format divergence risk.)

### 5. `GeminiConfig`: `authHeaders`, `timeout`, `maxRetries` are dead — low

`config/GeminiConfig.scala:29-32` returns exactly the two headers that `GeminiClientImpl.geminiAuthRequest` (PR 2, `GeminiClient.scala:35-38`) hardcodes independently, and `GeminiConfigSpec.scala:22` asserts the unused value — so the test passes while the two can silently diverge. `timeout`/`maxRetries` are documented in the scaladoc but never applied to the backend.

This is pre-existing rot in `core` (see global remark 4), but the cheapest in-PR fix is to build `geminiAuthRequest` from `config.authHeaders` so the abstraction and its test become load-bearing.

### 6. Minor

- `requests/InteractionRequest.scala:18` — `responseFormat.exists(_.isInstanceOf[ResponseFormat.JsonSchema])`; a match over the sealed ADT gets exhaustiveness checking if a third case appears.
- `models/GeminiModel.scala` has zero references in production code or tests — it appears only in `docs/gemini/models-and-errors.md`. Contrast `ClaudeModel`, which earns its place via `ClaudeModel.modelSupportsStructuredOutput` and has a `ClaudeModelSpec`. Fine as a convenience catalogue, but it's untested and unused, so it will silently drift.
- `ToolSpec:14` uses a null-heavy `params` fixture with the comment "parameters must be preserved verbatim, including legitimate nulls" — but this codec writes `"parameters" -> parameters` verbatim by construction; the null-dropping risk lives in `dropNullsPreservingToolParameters` (PR 2). The test is fine, the comment points at the wrong place.

---

## Global remarks (apply across the stack)

**G1 — `core`'s `asJson_parseErrors` default is wrong, and Claude is live-buggy because of it.**
`core/http/ResponseHandlers.scala:25-43` maps `asString`'s `Left(body)` — i.e. every non-2xx response — to `deserializationException`, so `mapErrorToException` is **never reached** for HTTP errors. OpenAI overrides it (`OpenAIJson.scala:31-38`), and now Gemini overrides it too (PR 2, `GeminiClient.scala:72-82`) with a comment explaining the default is wrong. **Claude does not override it** — so `ClaudeClientImpl.mapErrorToException` and its whole `authentication_error`/`rate_limit_error` dispatch are dead code for non-streaming calls, and a Claude 401 surfaces as `DeserializationClaudeException`.

Two providers working around the same broken default is the signal. Fix `ResponseHandlers.asJson_parseErrors` in core to the `asStringAlways` + `metadata.isSuccess` form, delete both overrides, and Claude's error mapping starts working. Net ~-20 lines. Doesn't have to be in this stack, but it should be a follow-up issue.

**G2 — `dropNullsPreserving*` is now the third copy of one algorithm, and it is incomplete in all of them.**
`GeminiClient.scala:145-169` is a verbatim copy of `ClaudeClient.scala:166-190` (only the field name differs); `OpenAIJson.scala:62-153` is a generalised third variant. See the detailed correctness finding on PR 2 — the point here is that a `core` helper (`CirceHelpers` already has `mergeExtraBody(fieldName)` in exactly this shape) would let both Claude and Gemini become one-liners with the ~6-line rationale comment living once, and would make it natural to preserve *all* verbatim-JSON paths rather than just `tools[*]`.

**G3 — `ensureObjectType` belongs in `core`.**
`gemini/agent/GeminiAgent.scala:42-48` is character-identical to `claude/agent/ClaudeAgent.scala:39-45`. It normalises `AgentTool.rawJsonSchema` — which `core` owns — so this is provider-independent MCP/JSON-Schema logic sitting in two provider modules. `core/agent/AgentTool.scala` is the home. ~8 lines, two call sites.

**G4 — dead abstractions in `core` that Gemini now instantiates a third time.**
- `core/client/AIClient.scala` — zero implementers repo-wide. Not Gemini's fault; should be deleted.
- `AIClientConfig.authHeaders` — implemented by all three configs, called by none.
- `AIClientConfig.timeout` / `maxRetries` — never read by any client; only asserted in config unit tests.

Worth a cleanup pass; pointing at this PR for it would be unfair, but it's now 3× instead of 2×.

**G5 — SSE adapter triplication is now 15 near-identical files.** Each Gemini streaming file is ~90% identical to its Claude sibling, which was already ~85% identical to the OpenAI one. This is the repo's established pattern and blocking a new provider on refactoring it would be the wrong trade — but the marginal cost of a 4th provider is now ~300 lines, so it's worth recording as debt. (A per-effect-module generic `parseSse[F, E, A](decode, mkError, doneSentinel)` would work; it can't live in `core`, which doesn't depend on fs2/zio/pekko.)

**G7 — is there a provider-neutral `Tool` this module should have reused? Not today, and mostly for good reasons.**
The only sttp-ai-wide tool abstraction is `core/agent/AgentTool.scala` — but that's a *callable* tool (name, description, schema, codec, `execute`), the input side of the agent loop. Each backend converts it to its own wire type, and `GeminiAgentBackend.convertTool` does that correctly, including using core's `rawJsonSchema` passthrough contract.

There is no provider-neutral **wire-level** `Tool`, and the three wire shapes genuinely differ: Claude `{"name","description","input_schema"}` + `cache_control` on every tool; Gemini `{"type":"function","name","description","parameters"}`; OpenAI chat `{"type":"function","function":{…}}` with a `strict` flag — and OpenAI has three separate `Tool` ADTs (chat/responses/assistants) already. The built-ins don't line up either (Claude `WebSearch` with domain lists + user location vs Gemini `GoogleSearch`/`CodeExecution`).

What's actually common is the declaration triple `(name, description, parameters: Json)`. A `core.agent.FunctionDeclaration` that each provider's sealed `Tool` wraps would save ~5 lines per provider and add a level of nesting to every codec — marginal. The part that *is* clearly worth lifting is `ensureObjectType` (G3), which is provider-neutral JSON-Schema normalisation of a `core`-owned value.

One note in this PR's favour: `Tool.Function.parameters: Json` (raw only) is a simplification over Claude, which carries both a typed `ToolInputSchema`/`PropertySchema` builder *and* `CustomRaw` — the typed one can't express nested objects or array item types, which is why `CustomRaw` had to be added later. Not repeating that here is the right call.

**G6 — `generated-docs/` staleness is expected, but has a user-visible consequence.** `git log -- generated-docs/` shows only release-driven commits, so leaving it untouched here matches practice. Just note that `.readthedocs.yaml` builds from `generated-docs/out/conf.py`, so the Gemini pages won't appear on sttp-ai.softwaremill.com until the next release regenerates it.

PR #538, opened 2026-07-27
Gemini/02 clients

flsh86:


adamw:
## Review — PR 2/6 (clients + exceptions)

Checked and correct: status→exception dispatch including the raw-body fallback when the error envelope doesn't parse (`GeminiClient.scala:48-65`); the `asJson_parseErrors` override that restores status dispatch for non-2xx (well spotted — see global remark G1 on #537, Claude is live-buggy for want of the same override, and you added a test pinning it); URI construction (`addPath` normalises trailing slashes on custom base URLs); auth headers; the codec init ordering.

### 1. Null-stripping corrupts every verbatim-`Json` field except `tools[].parameters` — high

`serializeInteractionRequest` (`GeminiClient.scala:142-169`) applies `deepDropNullValues` to the whole request and only splices `tools[*].parameters` back. The other raw-JSON fields are silently rewritten. Verified against the built module:

```scala
// response_format
InteractionRequest.simple("m","hi").copy(responseFormat = Some(ResponseFormat.JsonSchema(
  parse("""{"type":"object","properties":{"level":{"enum":["low","high",null],"default":null}}}""")))
→ {"model":"m","input":"hi","response_format":{"type":"object","properties":{"level":{"enum":["low","high"]}}}}
//                                                                                   ^ null gone, "default" gone

// replayed conversation steps
Step.FunctionCall("c1","get_weather", {"city":"Paris","unit":null})
Step.FunctionResult("c1","get_weather", {"temp":20,"error":null})
→ ..."arguments":{"city":"Paris"}..."result":{"temp":20}...
```

Two problems:

- `ResponseFormat.JsonSchema`'s scaladoc (`models/ResponseFormat.scala:14`) promises the schema is "sent to the API exactly as given". It isn't. The *same* schema passed as a tool's `parameters` **is** preserved (asserted by `GeminiClientSerializationSpec:40-48`), so the two paths disagree — and the tool path is the one that's tested.
- `Step.FunctionCall.arguments` / `Step.FunctionResult.result` inside `InteractionInput.StepsInput` is *conversation data*, not config. Replaying a call the model actually made as `{"city":"Paris","unit":null}` sends `{"city":"Paris"}` — the model is shown something different from what happened. (The agent backend in #539 wraps tool results with `Json.fromString`, so it's insulated; a direct API user replaying structured steps is not.)

Fix: preserve a set of paths (`response_format`, `tools[*].parameters`, `input[*].arguments`, `input[*].result`) rather than special-casing tools — or drop nulls in the derived-encoder layer instead of on the finished document. See global remark G2 on #537: this is the third copy of this algorithm, so it's a good moment to lift it into `core` and generalise the path list once.

### 2. The comment justifying the splice is factually wrong — medium

`GeminiClient.scala:139`: *"`deepDropNullValues` never removes array elements, only null values"*. It does remove null array elements — verified with circe 0.14.16:

```
{"a":[1,null,2],"b":{"c":null,"d":1},"e":["low","high",null]}
  .deepDropNullValues  →  {"a":[1,2],"b":{"d":1},"e":["low","high"]}
```

Claude's version of this comment says "never removes *object* elements from an array" — which is the correct and load-bearing claim (tool entries are objects, never null, so index alignment holds). The rewording here dropped the qualifier and inverted the meaning, and the wrong reason is exactly what makes the `"enum": [..., null]` case in #1 above look safe.

### 3. Parsed error `code` is decoded and then discarded — low/medium

`GeminiClient.scala:56-63` builds every exception with `code = None`, so `ErrorDetail.code` — including the string/number normalisation carefully hand-written at `GeminiManualCodecs.scala:15-33` — never reaches the user; `AIException.code` is always empty for Gemini. Either pass `parsed.error.code` through or drop the normalisation.

### 4. Test gaps on the two hand-written response handlers — low/medium

Both overrides exist precisely because the core defaults were wrong, but only the happy paths are covered:

- **Malformed / non-JSON error body** — `GeminiClient.scala:52` falls back to the raw body as the message. An HTML 503 page or an empty body takes this branch; untested. Same for `{"error":{"code":400}}` with no `message` (`errorDetailCodec` requires `message`).
- **Malformed 200 body → `DeserializationGeminiException`** (`GeminiClient.scala:77-80`) — `GeminiSyncClientSpec:50` looks like it covers this but doesn't; that exception comes from `GeminiSyncClient.scala:41`, a different path.
- `asUnit_parseErrors` (`GeminiClient.scala:84-90`), a *second* hand-written handler, has no unit coverage at all.
- `GeminiUris.interaction(id)` / `.cancel(id)`: `GeminiClientSerializationSpec:26` only verifies the POST. `cancelInteraction` has **no** coverage anywhere, not even integration; `deleteInteraction` only in the (CI-skipped) integration spec.
- `createInteractionAs`'s schema-injection branch (`GeminiSyncClient.scala:29-34`) — `GeminiSyncClientSpec:44` asserts only the decoded result, never the outgoing request, so neither the "derive schema" nor the "leave existing one alone" branch is verified. Claude tests exactly this (`ClaudeSyncClientSpec:78`).
- `createInteractionAsBinaryStream` (`GeminiClient.scala:117`) is a separate copy of the `stream = true` logic; only the `InputStream` variant is asserted (`GeminiClientSerializationSpec:82`).

### 5. Minor

- `GeminiClientSerializationSpec:37` — `body should not include "null"` is brittle; it fails on any legitimate payload containing the substring (a tool named `nullable`, a prompt mentioning "null"). Claude's equivalent parses and asserts key absence: `json.hcursor.downField("temperature").succeeded shouldBe false`. Same applies to the raw string matches in `GeminiAgentBackendSpec:92,113,114`.
- Test overlap: `GeminiClientSerializationSpec:50` (7-status table + a message assertion), `:64` (re-asserts `RateLimitException` + message, new info is only "string `code` decodes"), `:73` (re-asserts 401 + message) and `GeminiSyncClientSpec:38` assert 401 → `AuthenticationException` three times between them. `:73` is worth keeping — it's the only test of the `asJson_parseErrors` override wiring — the other two can shed their redundant assertions.
- `GeminiClientImpl` (`:31`) and `GeminiUris` (`:172`) are public; `private[gemini]` still compiles, since the only other caller lives in `sttp.ai.gemini.unit`. Claude/OpenAI leak these too, so this is following convention — but nothing stops Gemini being the one that doesn't.
- `GeminiSyncClient` repeats the `send(backend).body match { Left => throw; Right => v }` unwrap 4×; a one-line private helper would do.
- `GeminiSyncClient.scala:32` inlines `TapirSchemaToJsonSchema(..., markOptionsAsNullable = true)` where Claude wraps it in a named constructor (`OutputFormat.JsonSchema.withTapirSchema[T]`). A `ResponseFormat.JsonSchema.withTapirSchema[T]` would read better and match the sibling module.

PR #539, opened 2026-07-27
Gemini/03 agent

flsh86:


adamw:
## Review — PR 3/6 (agent-loop backend)

Conversation state is correct: the response contains only newly generated steps (confirmed by the live fixture in `InteractionResponseSpec`), so `functionCalls` collecting across all steps doesn't resurrect prior calls; tool-call/result pairing via `callId` is sound; nothing is lost or duplicated across iterations. `AgentResponse` has no usage field in the core contract, so discarding `response.usage` matches the Claude and OpenAI backends. The stateless-replay test (`GeminiAgentBackendSpec:90`) is a genuinely valuable invariant to pin.

### 1. A malformed `function_call` step silently ends the loop with an empty answer — high (root cause is in #537)

Because the `Step` codec falls back to `Step.Unknown` on *any* decoding failure (see finding 1 on #537), and `Step.FunctionCall(id, name, arguments)` has no defaults, a response step missing `id` **or** `arguments` decodes to `Unknown` and disappears. Verified:

| response step | `functionCalls` |
|---|---|
| `{"type":"function_call","id":"c1","name":"now"}` (no `arguments`) | `List()` |
| `{"type":"function_call","name":"now","arguments":{}}` (no `id`) | `List()` |

Path from there: `GeminiAgent.scala:89` sees zero tool calls → `mapStopReason:101` sees `Completed && !hasToolCalls` → `StopReason.EndTurn` → `Agent`'s empty-tool-calls branch → `AgentResult(finalAnswer = "", finishReason = NaturalStop)`. The tool is never invoked, nothing is logged or raised, and the caller gets a successful empty run.

This is most likely to bite exactly the case the backend goes out of its way to support: `ensureObjectType` (`:38-48`) exists to normalise no-argument tool schemas, and a no-argument tool is the most plausible source of an omitted `arguments`.

Related, same root cause: `"arguments": null` decodes fine and yields `ToolCall(input = "null")` at `:89`, which then fails `Agent.executeTool`'s decode and reaches the model as a parse-error string rather than an empty-object call.

Fixing the discriminator-gated fallback in #537 addresses this; defaults on `Step.FunctionCall` would be belt-and-braces.

### 2. `Cancelled` is reported as a clean success — medium

`GeminiAgent.scala:86` errors only on `InteractionStatus.Failed`. `Cancelled` falls through `mapStopReason`'s catch-all (`:106`) to `StopReason.Other("cancelled")`, which `Agent.loop` doesn't special-case, so with no tool calls it returns `FinishReason.NaturalStop` with `finalAnswer = ""`. `Cancelled` should join `Failed` in the error branch.

### 3. The `Failed` error message is permanently uninformative — medium

```scala
store = Some(false)                                                                   // :81
monad.error(new RuntimeException(s"Gemini interaction ${response.id.getOrElse("<unstored>")} failed"))  // :87
```

Since the backend always sends `store = false`, the API never returns an `id` (documented in the PR description and confirmed by `InteractionResponseSpec:50-53`) — so this message is *always* `"Gemini interaction <unstored> failed"`, and the rest of the response is discarded. Include `response.status` / `model` / `outputText` instead.

### 4. Please confirm: `Step.Thought` is never replayed — needs a live check

`buildSteps` (`:50-68`) reconstructs request steps purely from `ConversationHistory`, which has no slot for thought steps. Live responses do contain them with an opaque `signature` — the verbatim fixture at `InteractionResponseSpec:53` has one. So from iteration 1 onward, a tool-using loop replays `user_input, model_output, function_call, function_result` with the preceding `thought` step and its signature dropped.

The information loss is certain; whether it *matters* depends on whether the Interactions API enforces the Gemini 3 rule that thought signatures must be echoed alongside replayed function calls. Since you have a live key and the agent integration spec exercises multi-iteration tool use, a quick check would settle it. If it does matter, there's currently no mechanism in `ConversationHistory` or the backend that could carry the signature.

### 5. `ensureObjectType` is character-identical to `ClaudeAgentBackend`'s — medium

`GeminiAgent.scala:42-48` vs `ClaudeAgent.scala:39-45`. It normalises `AgentTool.rawJsonSchema`, which `core` owns, so this is provider-independent MCP/JSON-Schema logic living in two provider modules. `core/agent/AgentTool.scala` is the home. See global remark G3 on #537.

### 6. Minor

- `:58` — `parseJson(tc.input).fold(throw _, identity)` runs in the *eager* part of `sendRequest`, before any `F` is constructed, so for `F = IO`/`Task`/`ZIO` a `ParsingFailure` escapes the effect's error channel entirely and `ExceptionHandler` never sees it. In the agent loop `tc.input` always comes from `fc.arguments.noSpaces` so it can't fire today, but a user-built `ConversationHistory` can. `ClaudeAgentBackend` has the same shape, so this is baseline, not a regression — just cheap to fix in new code by returning `Either` from `buildSteps`.
- `:93-94` — `monad.error(new RuntimeException(s"Gemini API error: ${error.getMessage}"))` throws away an already-typed `GeminiException` carrying status, cause and type, so callers can't distinguish `RateLimitException` from `InvalidRequestException` to decide on a retry. OpenAI and Claude do the same, so again baseline — but `monad.error(error)` is strictly better and free.

### 7. Test gaps

Every backend in `GeminiAgentBackendSpec` is built with `systemPrompt = None, responseSchema = None` (`:33`), so these are unexercised:

- `systemPrompt` → `system_instruction` (`GeminiAgent.scala:78`)
- `responseSchema` → `responseFormat`, including the `deepDropNullValues` at `:28-29`
- `ensureObjectType`'s **object-missing-`type`** branch (`:46`) — the `{}` no-argument MCP tool the comment calls out. The boolean branch (`:49`) and passthrough (`:36`) are covered.
- `mapStopReason`: `Incomplete`/`BudgetExceeded` → `MaxTokens` and the `other` catch-all (`:104-106`)
- the HTTP-error path (`:93`) — only the `failed`-status path is tested (`:142`)
- `ConversationEntry.IterationMarker` → user text step (`:66`)

Extending `captureRequestBody` to take systemPrompt/responseSchema covers the first two on the wire; one parameterised case list covers the stop reasons.

PR #540, opened 2026-07-27
Gemini/04 streaming

flsh86:


adamw:
## Review — PR 4/6 (streaming extensions)

**Correctness: clean.** All five extensions are faithful ports of their Claude siblings in the same directories — identical SSE pipeline (`*ServerSentEvents.parse` → filter → decode), identical blank-data and `[DONE]` filtering, identical `ResponseMetadata` capture at the response-mapping boundary, and the terminal-error semantics each effect system already uses (`.rethrow` for fs2, `collectZIO` for zio, throw-into-stream for akka/pekko, `Either` in the flow for ox). The ox variant correctly goes via `createInteractionAsInputStream`, which sets `stream = true`. No divergence that is a bug.

**build.sbt wiring: correct and complete.** All five modules gain `gemini % "compile->compile;test->test"` alongside openai and claude, and `gemini`'s `jvmPlatform(scalaVersions = scala3 ++ scala2)` covers both the 2.13-only `akka` row and the 3-only `ox` row. Verified by compiling `gemini`, `gemini3`, `fs2`, `fs23`, `zio3`, `akka`, `pekko3`, `ox3`.

Two things to fix:

### 1. `DoneEvent` is copy-pasted into all five files — medium

`private val DoneEvent = "[DONE]"` plus its identical 2-line justification comment appears in `GeminiFs2Streaming.scala:21`, `GeminiZioStreaming.scala:22`, `GeminiPekkoStreaming.scala:22`, `GeminiAkkaStreaming.scala:22`, `GeminiOxStreaming.scala:20`.

Both other providers define it exactly once, in the module that owns the protocol, and import it: `claude/.../responses/MessageStreamResponse.scala:96`, `openai/.../ChatChunkRequestResponseData.scala:39-40`. A wire-protocol sentinel isn't a streaming-adapter concern. Move it to the `InteractionStreamEvent` companion (`gemini/.../responses/InteractionStreamEvent.scala`) and import in the five adapters — ~10 lines, entirely within this PR.

### 2. No tests for any of the five extensions — medium

Baseline: `claude` ships `streaming/ox/src/test/scala/sttp/ai/claude/streaming/ox/ClaudeOxClientSpec.scala`, and `openai` has a client spec in *every* effect module. Gemini adds five files and zero tests.

The untested behaviours aren't trivial, and three of them differ per module:

- The `[DONE]` sentinel is explicitly speculative per its own comment ("does not document a terminating sentinel… defensively") — exactly the kind of thing a test should pin so it isn't quietly deleted later.
- Malformed frame → `DeserializationGeminiException`, surfaced as a `Left` element in ox but as a **stream failure** in fs2 (`.rethrow`) and zio (`collectZIO`). Three different contracts, none asserted.
- Non-2xx → status-mapped exception via `asStreamUnsafe_parseErrors`. `ClaudeOxClientSpec:39-73` is a status-code table doing precisely this and would port almost verbatim.

Porting `ClaudeOxClientSpec` (ox at minimum, ideally fs2 too since its `.rethrow` semantics differ) would close most of this.

PR #541, opened 2026-07-27
Gemini/05 Test(gemini): add integration specs for client and agent loop

flsh86:
Add GeminiIntegrationSpec (client CRUD/structured-output) and GeminiAgentIntegrationSpec (agent loop, extending AgentIntegrationSpecBase), and wire *GeminiIntegrationSpec/*GeminiAgentIntegrationSpec into run-integration-tests.sh alongside a GEMINI_API_KEY guard.

### DIFF
--- a/streaming/akka/src/main/scala/sttp/ai/gemini/streaming/akka/GeminiAkkaStreaming.scala
+++ b/streaming/akka/src/main/scala/sttp/ai/gemini/streaming/akka/GeminiAkkaStreaming.scala
@@ -12,7 +12,6 @@ import sttp.client4.akkahttp.AkkaHttpServerSentEvents
 import sttp.model.ResponseMetadata
 import sttp.model.sse.ServerSentEvent
 import io.circe.parser.decode
-import sttp.ai.gemini.json.GeminiManualCodecs._
 import sttp.ai.gemini.json.GeminiDerivedCodecs._
 
 object GeminiAkkaStreaming {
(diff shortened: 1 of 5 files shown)


adamw:
## Review — PR 5/6 (integration specs)

`run-integration-tests.sh` is correct: the changes mirror the existing OpenAI/Anthropic entries exactly (flag variable, presence check, usage text, final summary condition), both spec globs are added to `testOnly`, `*GeminiIntegrationSpec` does not also match `GeminiAgentIntegrationSpec` so there's no double-run, and the `.env` sourcing loop picks up `GEMINI_API_KEY` with no change needed. `GeminiAgentIntegrationSpec` is a faithful use of `AgentIntegrationSpecBase`. Auto-skip via `cancel(...)` and client cleanup both match `ClaudeIntegrationSpec`.

Three gaps against the project's own convention:

### 1. No output cap — the cost convention isn't followed

`ClaudeIntegrationSpec` passes `maxTokens = 5/10/20` on every request (`:102,:129,:166`). Neither Gemini spec sets `GenerationConfig.maxOutputTokens` on any request, so output length is unbounded on a thinking-capable model — and thought tokens are billed. Add `generationConfig = Some(GenerationConfig(maxOutputTokens = Some(...)))`.

### 2. No error-path test

`ClaudeIntegrationSpec:290` hits the real API with an invalid key and asserts the mapped exception. That's the only end-to-end validation that `mapErrorToException` matches what the server actually returns — Gemini's status dispatch is currently validated solely against synthetic bodies. Worth adding especially because this test needs no API key, so it runs in CI on every build.

### 3. Model id inconsistency

Integration specs hardcode `"gemini-3.5-flash-lite"` (`GeminiIntegrationSpec:26`, `GeminiAgentIntegrationSpec:13`) while unit specs use `"gemini-2.5-flash-lite"`. Per your own note in `docs/gemini/models-and-errors.md:33`, 2.5-lite returns 404 for new API users — so the unit-spec value is a name that can't be requested. Harmless in stubs, but it's a trap for anyone copying it. Neither uses the `GeminiModel` constants, which is why those constants have no test coverage at all (see the note on #537).

Optional: no tool-calling test against the real API (Claude has one); the agent spec covers it indirectly, so this is a judgement call.

PR #542, opened 2026-07-28
Gemini/06 docs

### LINKED ISSUE #536: Documentation review
* main page: an overview of features: api access, agent loop, so that ppl know what’s possible with the project
* quickstart: apart from deps, also include a starting code snippet
* openai/basics.html:
    * link is not a link
    * which openai features are available
    * sth above basic usage? maybe the title of the page should be "OpenAI API basics" or sth like that
    * Token limits on GPT-5 / reasoning models - should be a section
* openai/streaming.html
    * sections for fs2/ox
    * at least list the deps for other libs
    * whenever tapir, fs2 etc. is mentioned for the first time -> link to project
* openai/structured-outputs.html
    * the flow is weird and unclear: first its described how to derive schema via Tapir, then manually, then via Tapir, but differently (?), then again with Tapir?
    * it should be structured as: what problem does it solve - what is needed (json schema) - how to achieve it the easiest way - how to customise - how to do manually
* openai/compatible-apis.html
    * list ollama, grok, openrouter in the page title
    * better sectioning (top-level, not nested)
* claude/basics.html
    * page title could be better, but the features are nicely listed - we should have the same in openai
    * claude/messages.html - lacking information "what is this for"
    * claude/structured-outputs.html - can the information about different options for deriving structured outputs schemas be somehow common for openai & claude?
    * claude/tool-calling.html - why is there no tool calling section for openai? or maybe this is common for both?
    * claude/models-and-errors.html - seems like a catch-all page. Sync client should be the default one that we propose. Rework this or split into separate pages
* agents/tools.html
    * schema derivation could be shared with schema discussion for structured outputs?
* other/backends.html
    * the "other backends" seems relevant only for agent loop?
    * effect systems is only for openai?

flsh86:


adamw:
## Review — PR 6/6 (docs)

I ran `sbt compileDocumentation` on the tip of the stack: **all 15 `mdoc:compile-only` blocks in `docs/gemini/*.md` compile** (`Compiled in 21s (0 errors)`). Types, method names, parameter names/arity, import paths, the `derives io.circe.Codec.AsObject, Schema` requirements for `createInteractionAs[T]` and `AgentTool.fromFunction`, and both streaming extension imports are all correct. No signature mismatches.

Worth calling out: the Gemini docs are *better* covered than the existing Claude ones — every executable example here is mdoc-checked, whereas `docs/claude/{messages,models-and-errors,streaming,tool-calling}.md` have zero checked blocks. Also, the `claude-streaming-fs2` → `fs2` coordinate rename across README/quickstart/Readme-prefix is a genuine bug fix; the old coordinates never existed. Nav wiring (`docs/index.md:38-49`) all resolves, and `build.sbt:261` correctly adds `gemini` to the docs `dependsOn`.

### 1. `models-and-errors.md:29` contradicts `:33`, and `:29` is wrong

`:29` — "`Gemini25FlashLite` (`gemini-2.5-flash-lite`) — fastest and cheapest, **good for integration tests**". `:33` — "the 2.5-generation `-lite` models … are no longer available to new Gemini API users — requesting them returns a 404". Your own integration specs use `gemini-3.5-flash-lite`. Drop the "good for integration tests" clause.

### 2. `interactions.md:102-111` — the `Step` listing is incomplete, and it's in an unchecked block

The real `models/Step.scala` has two more cases:

```scala
case class Thought(signature: Option[String] = None, content: Option[List[Content]] = None) extends Step  // :17
case class Unknown(raw: io.circe.Json) extends Step                                                        // :20
```

Not academic — `Thought` steps come back from the live API; the recorded fixture at `InteractionResponseSpec:53` contains `{"signature":"...","type":"thought"}`. A reader writing an exhaustive match over the four documented cases gets a `MatchError` at runtime. (It's a plain ```` ```scala ```` fence, so mdoc can't catch the drift.)

### 3. `CLAUDE.md` still points at uPickle/SnakePickle in three places

`:71` was correctly fixed to "All use **circe with snake_case configuration**", but these remain:

- `:128` — "**Naming**: Snake_case for JSON fields (handled by SnakePickle)"
- `:155` — "JSON serialization debugging (test uPickle behavior)"
- `:166` — `import sttp.ai.claude.json.SnakePickle._` in the scratch-file template

`grep -r SnakePickle --include=*.scala` returns zero hits repo-wide; `:166` is a copy-pasteable snippet that cannot compile.

### 4. `CLAUDE.md:87` — stale streaming API summary

"**Pattern:** Extension methods add `.parseSSE` + `.parseClaudeStreamResponse`/`.parseOpenAIStreamResponse`" — none of those three identifiers exist anywhere in the codebase. The actual API is `createStreamedInteraction` / `createStreamedMessage` / etc. The PR rewrote the streaming table immediately above this line and left the summary underneath it.

### 5. `README.md:15,17` weren't updated, though the identical lines in `docs/index.md:6,8` were

README still reads "…wrapper for OpenAI, Claude (Anthropic), and OpenAI-compatible APIs. Use the power of ChatGPT and Claude inside your code!" while the same file now carries a Gemini badge (`:9`) and a Gemini quickstart (`:48`).

### 6. `docs/conf.py:27` — not in this PR, but should be

`llms_txt_summary = "Scala client for OpenAI, Claude (Anthropic), and OpenAI-compatible APIs"` is published as the summary of the generated `llms.txt`, so the machine-readable site summary omits Gemini.

### 7. `CLAUDE.md` reference sections not extended

Gemini was added to the command list (`:33`) but not to `:137` (integration-test env vars — still only `OPENAI_API_KEY`/`ANTHROPIC_API_KEY`) or `:143-146` (client tables — still only `OpenAISyncClient`/`ClaudeSyncClient`, `OpenAI`/`ClaudeClient`).

### 8. Minor

- `CLAUDE.md:15` / `:139` say "Scala 2.13.16 and Scala 3.3.6"; `build.sbt:6-7` is `2.13.18` / `3.3.8`. Pre-existing, but `:15` is inside a bullet list this PR edits.
- `gemini/basics.md:15` says "Cross-platform — Scala 2.13 and Scala 3" but `build.sbt:105-107` also gives `gemini` a `nativePlatform(scalaVersions = scala3)` row; Scala Native goes unmentioned.
- `interactions.md:44-52` is the only executable Gemini snippet not mdoc-checked (real API usage with `.copy(store = Some(true))`). Adding imports + `mdoc:compile-only` would make coverage complete; the other three plain blocks are type declarations, matching the `docs/claude/*.md` convention.
- FYI: `.readthedocs.yaml` builds from `generated-docs/out/`, which is only regenerated on release — so the Gemini pages won't appear on sttp-ai.softwaremill.com until the next release. That matches established practice (`git log -- generated-docs/` shows only release commits), just noting the lag.

PR #544, opened 2026-07-29
Docs/01 foundation

flsh86:
- Add a "What can you do with sttp-ai?" feature overview to the main docs page (`index.md`), linking each capability (direct API access, compatible providers, streaming, structured outputs, tool calling, agent loop) to the page documenting it
- Add starting code snippets to the quickstart: a minimal, mdoc-verified sync-client "hello" example for each provider (OpenAI, Claude, Gemini), and add the streaming-modules dependency list to the OpenAI section (previously only Claude/Gemini had it)
- Sweep stale content: `sttp-openai` → `sttp-ai` project-name references, the ChatProxy example link (old repo name), and outdated Claude model ids (`claude-3-*` → current 4.5 ids) across `agents/quickstart.md`, `claude/basics.md`, `claude/messages.md`, `claude/tool-calling.md`, and `openai/streaming.md`

PR #547, opened 2026-07-29
Docs/04 claude gemini agents

flsh86:
- Rewrite `claude/basics.md` sync-first: `ClaudeSyncClient` is the recommended default shown in Basic usage, the async `ClaudeClient` example moves to its own section, and the feature list now links to the page documenting each feature
- Add a "what is this for" intro to `claude/messages.md` and make all its snippets self-contained and mdoc-compiled (previously none were)
- Rework `claude/tool-calling.md` around the tool-result round trip (execute the tool, send `ToolResult` back, get the final answer) — the piece readers actually need, previously missing — plus an agent-loop pointer; all snippets now compile-checked
- Focus `claude/models-and-errors.md` to "Models and error handling": drop the stale Claude 3 model list and the OpenAI comparison table (basics covers key differences), move the sync client to basics. The error-handling snippet is fixed to use the real nested exception hierarchy (`ClaudeExceptions.ClaudeException.*`) — the old flat import never compiled
- Apply the same dissolution to `gemini/models-and-errors.md` and linkify the Gemini feature list, for cross-provider consistency
- Move `other/backends.md` → `agents/custom-backends.md`: its content is entirely agent-loop material (`AgentBackend[F[_]]`). The effect-system examples are now self-contained and compile-checked (they need `MonadError[F]` givens and `AgentTool.fromFunctionF` — the old snippets referenced undefined values and didn't compile), and note Gemini alongside OpenAI/Claude
- Add Gemini to the agent-loop quickstart (intro, feature list, and a `GeminiAgent` pointer)
- Fix wrong Claude model ids found in final review (Sonnet 4.5 is `claude-sonnet-4-5-20250929`; Opus 4.1 is `claude-opus-4-1-20250805`) plus small prose nits


docs/claude/structured-outputs.md
```diff
@@ -90,7 +90,7 @@ object StructuredOutputExample:
     )
 
     val request = MessageRequest
-      .simple("claude-sonnet-4-5-20250514", messages, 500)
+      .simple("claude-sonnet-4-5-20250929", messages, 500)
```
adamw: we're at Sonnet 5, probably needs updating

adamw:
Review of the whole docs stack (#544 → #545 → #546 → #547), posted here as the last PR in the series. Every API name, model id, and claim in the diffs was verified against `master` code and the final docs state on this branch. Overall verdict: **solid overhaul, all four PRs are mergeable** — the findings below are improvements, not blockers.

## Overview

The stack takes the docs from "per-page accretion" to a coherent structure: a capability-oriented landing page, runnable hello-world quickstarts per provider, one shared home for JSON-schema derivation, OpenAI pages brought up to Claude/Gemini structure, and sync-first provider basics with a proper tool-calling round trip for Claude. Converting previously-unchecked snippets to `mdoc:compile-only` is the biggest quality win — many of the old snippets (Claude error handling, messages, backends) never compiled.

**Correctness — verified, all good:**
- Every API referenced in prose and new snippets exists: `createChatCompletionAs`/`createMessageAs`/`createInteractionAs`, `Tool.Function.withSchema`, `Tool.CustomRaw`, `AgentTool.fromFunction`/`fromFunctionF`, `deriveResponseSchema`, `GeminiAgent.synchronous`, `Message.toolResult`, all `fromEnv` variants, and every exception subclass in the new error-handling snippet (including `TryAgain`, `ServiceUnavailableException`; the nested `ClaudeException.*` hierarchy matches the source, and the old flat import indeed never compiled).
- Final-state model ids are correct (`claude-sonnet-4-5-20250929`, `claude-haiku-4-5-20251001`, `claude-opus-4-1-20250805`, `claude-opus-4-5-20251101`); no stale `claude-3-*` or wrong ids remain outside `docs/plans/`.
- The Grok→Groq fix is right: `api.groq.com` is Groq, `llama-3.1-8b-instant` is a Groq model, and the `GROQ_API_KEY` env var fix (was `OPENAI_KEY`) is a real bug caught. The x.ai note is accurate.
- Dropping `.getOrElse("N/A")` on `displayName` is correct — it's `String`, not `Option`.
- Env var names match the code (`GEMINI_API_KEY`, `ANTHROPIC_API_KEY`).

## Findings, by importance

**1. `openai/structured-outputs.md` lost its lead-in (#545 + #547).** After #547, the "Typed responses with createChatCompletionAs[T]" heading drops straight into a 25-line code block. The sentence that used to explain the mechanism — schema derived via Tapir, set as `responseFormat` automatically, response parsed via circe — was deleted. One lead-in sentence before the code would restore the flow (the Claude/Gemini pages have one).

**2. The strict-mode section references a knob the reader hasn't seen yet.** "Normalization is applied only when `strict = true` is requested" — but `strict` is a field on `ResponseFormat.JsonSchema`, which is only introduced in the *following* section. Either move "Strict mode and schema normalization" after the lower-level section, or open it with "when you pass `strict = Some(true)` to `ResponseFormat.JsonSchema` / `withTapirSchema`". Also: the text never says *who* normalizes (this library does); and the ~110-column hard wraps inherited from the old blockquote should be reflowed.

**3. `other/json-schemas.md` snippets aren't standalone-runnable.** It's the only new page whose `mdoc` snippets have no `//> using dep` header — and `TapirSchemaToJsonSchema` needs `tapir-apispec-docs` (the deleted snippet pinned it explicitly). Since the docs promise "examples are runnable using scala-cli", add the dep lines. Otherwise this page is the best addition in the stack — it genuinely dissolves the 3-way duplication.

**4. Quickstart hello examples are inconsistent with each other (#544).** The Claude and Gemini examples `close()` their clients; the OpenAI one doesn't, though `OpenAISyncClient` has `close()`. OpenAI also prints the raw response object while the others print just the content. Being the very first code a reader sees, these three should be symmetrical.

**5. Duplication left behind (#546):** the "Client implementations" section at the bottom of `compatible-apis.md` now restates the new "Sync and async clients" section in `openai/basics.md`. Replace with a link.

**6. Pekko streaming could be compile-checked (#546).** The Akka/Pekko section is prose-only, justified by "Akka is Scala 2.13-only" — but Pekko supports Scala 3. If the pekko module can go on the docs classpath, a verified Pekko example would complete the set; the "consume the `Source` as in the fs2 and ZIO examples above" hand-wave is the weakest spot on an otherwise much-improved page.

**7. Minor:** the heading `## Typed responses with createChatCompletionAs[T]` lacks backticks around the method name, unlike sibling headings ("Lower-level: building `ResponseFormat.JsonSchema` yourself"). And in `claude/models-and-errors.md`, the note that `UnsupportedModelForStructuredOutputException` is "also under `ClaudeExceptions`" is accurate but could mislead — it extends `AIException` directly, not `ClaudeException`, so a `case e: ClaudeException` catch-all won't catch it; worth one clarifying clause.

## Advantages highlighting

The overhaul helps a lot here — the index "What can you do with sttp-ai?" list, the linkified per-provider feature lists, and instant hello-world snippets all make the library's breadth visible. What's still missing is a *differentiation* statement: nothing tells a reader why sttp-ai over openai-scala-client, langchain4j, or hand-rolled HTTP. The ingredients are all documented but never assembled: **three native provider APIs (not compatibility shims) in one library; bring-your-own effect system including direct-style Ox — unique in this space; schemas and tools derived from case classes via Tapir instead of hand-written JSON; an agent loop with typed tools/results plus MCP tool loading; Scala 2.13/3 and Scala Native (Gemini)**. A 3–4 sentence "Why sttp-ai?" block on `index.md` right after the feature list would close this, and #544 is the natural PR to carry it.

## Notes / follow-ups (non-blocking)

- #544 introduces the wrong id `claude-sonnet-4-5-20250514`, which only #547 fixes — fine if the stack merges promptly and in order, but fixing it at the source in #544 would keep intermediate master states clean.
- Out of scope for these PRs but found while verifying: the exception message in `ClaudeExceptions.scala:22` still cites the non-existent ids `claude-sonnet-4-1-20250514, claude-opus-4-1-20250514` — worth a code follow-up.
- `docs/plans/*.md` still carry `sttp-openai` naming; they're not in the toctree, so consider deleting them in a cleanup PR.
- Moving `other/backends.md` → `agents/custom-backends.md` (the right home) breaks any external links to the old published URL; if readthedocs redirects are configured for this project, add one.

PR #551, opened 2026-07-30
Azure OpenAI support

### LINKED ISSUE #143: Azure support
Add Azure support by letting the user specify the endpoint and an "api-key" header.
https://learn.microsoft.com/en-us/azure/ai-services/openai/chatgpt-quickstart?tabs=bash%2Cpython&pivots=rest-api
I'm willing to make that change if no one is working on it.

flsh86:
## What

Adds Azure OpenAI support to the `openai` module by making the authentication header configurable. Azure's API-key flow uses an `api-key: <key>` header instead of `Authorization: Bearer <key>`; combined with the already-supported custom base URL (query strings like `?api-version=...` are preserved), this is everything needed to use the client against Azure OpenAI deployment endpoints.

Closes the "Add Azure support by letting the user specify the endpoint and an api-key header" feature request.

## How

New `AuthScheme` ADT in `sttp.ai.openai`:

```scala
sealed trait AuthScheme
object AuthScheme {
  case object Bearer extends AuthScheme      // Authorization: Bearer <key> (default)
  case object AzureApiKey extends AuthScheme // api-key: <key>
}
```

Usage:

```scala
val client = OpenAISyncClient(
  authToken = azureKey,
  baseUrl = uri"https://my-resource.openai.azure.com/openai/deployments/gpt-4o?api-version=2024-10-21",
  authScheme = AuthScheme.AzureApiKey
)
```

- `OpenAI` gains a 4th defaulted constructor param `authScheme: AuthScheme = AuthScheme.Bearer`; the only behavioral branch is in `openAIAuthRequest`. Organization-header logic unchanged.
- `OpenAIConfig` gains an `authScheme` field (last position, defaulted); `authHeaders` and `OpenAI.apply(config)` honor it. `fromEnv` unchanged.
- `OpenAISyncClient` gains two new `apply` overloads taking `authScheme`. They take it as a **required** parameter (not defaulted) because Scala forbids default arguments on multiple overloaded alternatives — this is the one intentional deviation from "defaulted param" phrasing in the design.
- `customizeRequest` previously dropped `organization` when re-wrapping the client (latent bug); it now preserves all constructor state including `authScheme`.
- Streaming modules need no changes — they delegate to the `OpenAI` instance.
- Default behavior is fully backward compatible: all existing constructors/`apply`s keep sending `Authorization: Bearer`.

## Testing

- **Unit tests** (both Scala 2.13 and 3): header presence *and* absence pinned in both directions at three levels — raw request builder (`AuthSchemeSpec`), config-derived headers, and an end-to-end sync-client stub-backend test asserting on the actually-sent request headers (`SyncClientSpec`).
- **Integration test**: new `AzureOpenAIIntegrationSpec`, auto-skipped (canceled, not failed) unless `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_ENDPOINT` are set — same pattern as the OpenAI/Claude/Gemini specs. Wired into `run-integration-tests.sh` and documented in `INTEGRATION_TESTING.md`.
- `sbt compileDocumentation` compiles the new docs snippet.

## Docs

New "Azure OpenAI" section in `docs/openai/compatible-apis.md` covering the `api-key` scheme with a classic deployment URL, plus a note that Azure's newer `/openai/v1/` endpoint also works with the default Bearer scheme.

### DIFF
--- a/INTEGRATION_TESTING.md
+++ b/INTEGRATION_TESTING.md
@@ -23,6 +23,25 @@ export OPENAI_API_KEY=your-openai-api-key-here
 export ANTHROPIC_API_KEY=your-anthropic-api-key-here
 ```
 
+### Azure OpenAI
+
+To run the Azure OpenAI integration tests, set:
+
+```bash
+export AZURE_OPENAI_API_KEY=your-azure-api-key
+export AZURE_OPENAI_ENDPOINT="https://my-resource.openai.azure.com/openai/deployments/gpt-4o-mini?api-version=2024-10-21"
+# optional - deployment name sent in the request body (defaults to gpt-4o-mini)
+export AZURE_OPENAI_DEPLOYMENT=gpt-4o-mini
+```
+
+`AZURE_OPENAI_ENDPOINT` must be the full base URL including the deployment path and `api-version` query. Then run:
+
+```bash
+sbt "testOnly *AzureOpenAIIntegrationSpec"
+```
+
+If the variables are not set, the tests are skipped (not failed).
+
 ### Running Tests
 
 #### Run all integration tests:
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,14 +4,14 @@ sttp is a family of Scala HTTP-related projects, and currently includes:
 
 * [sttp client](https://github.com/softwaremill/sttp): the Scala HTTP client you always wanted!
 * [sttp tapir](https://github.com/softwaremill/tapir): typed API descriptions
-* sttp ai: this project. Non-official Scala client wrapper for OpenAI, Claude (Anthropic), Gemini (Google), and OpenAI-compatible APIs (e.g. Ollama, Grok, Groq, OpenRouter). Use the power of ChatGPT, Claude, and Gemini inside your code!
+* sttp ai: this project. Non-official Scala client wrapper for OpenAI, Claude (Anthropic), Gemini (Google), and OpenAI-compatible APIs (e.g. Azure OpenAI, Ollama, Grok, Groq, OpenRouter). Use the power of ChatGPT, Claude, and Gemini inside your code!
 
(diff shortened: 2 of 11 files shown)


adamw:
## Review

The approach is sound: a minimal `AuthScheme` ADT with the single behavioral branch in `openAIAuthRequest`, defaulted everywhere so existing code is source-compatible. Verified locally on this branch: `AuthSchemeSpec` and `SyncClientSpec` pass on both Scala 2.13 and 3 (29 tests), and scalafmt checks are clean. The `customizeRequest` fix (previously dropped `organization` when re-wrapping the client) is a good catch.

### Findings

1. **Duplicated auth-header logic** — the `AuthScheme` match appears twice with the same header names: `OpenAI.openAIAuthRequest` and `OpenAIConfig.authHeaders`. If the two ever drift, config-built and directly-built clients would authenticate differently. Consider a single source of truth, e.g. a method on `AuthScheme`:
   ```scala
   def authHeader(apiKey: String): (String, String)
   ```
   and using it from both places (`basicRequest.header(name, value)` on the client side).

2. **Binary compatibility** — adding a constructor parameter to `OpenAI` and a field to `OpenAIConfig` breaks binary compatibility with 0.5.x (old `<init>`/`apply`/`copy` signatures disappear). There's no MiMa in the build and the library is 0.x, so this is likely acceptable — just flagging it so it's a conscious choice for the release notes.

3. **Minor, in `SyncClientSpec`** — the stub returns a valid models-list response, so the `scala.util.Try(...)` wrapper around `getModels` is unnecessary; a plain call would also fail louder if the stub setup ever breaks. Not blocking.

Everything else follows existing conventions: the new `OpenAISyncClient.apply` overloads mirror the existing ones (`closeClient = false` when a backend is passed), the integration spec matches the `cancel`-when-unset pattern of the other providers, test placement matches the existing layout, and the docs snippet follows the style of the other compatible-API sections. The doc note about the newer `/openai/v1/` endpoint accepting Bearer auth is accurate and useful.

PR #552, opened 2026-07-30
Feature/reasoning content

### LINKED ISSUE #399: support reasoning content
There's reasoning content from models now, but `ChatChunkResponse` doesn't show it. Also `AssistantMessange` doesn't have the field

flsh86:


adamw:
## Review

**Overview:** adds `reasoningContent: Option[String]` to the chat response `Message` and streaming `Delta`, populated from `reasoning_content` (DeepSeek/Qwen/vLLM/Ollama) with a fallback from the `reasoning` spelling (OpenRouter/Groq/xAI). The fallback is implemented as a `Decoder#prepare` normalization step wrapped around the derived codecs, so every client and all streaming effect systems pick it up for free — good placement.

Verified locally on this branch: `ChatDataSpec` and `ChatChunkDataSpec` pass on both Scala 2.13 and 3, scalafmt checks are clean, and the full openai unit suite passes. (One `OpenAIAgentIntegrationSpec` test failed against the real API — `MaxIterations was not equal to NaturalStop` — but it fails nondeterministically across runs and is unrelated to this decode-only change.)

Test coverage is thorough: both spellings, precedence, null-`reasoning_content` fallback, non-string `reasoning` tolerance, absence, and the encode side are all pinned, for both the response message and the streaming delta.

### Findings

1. **Asymmetric strictness for `reasoning_content`** — commit ab96af4 deliberately hardened the decoder so a non-string `reasoning` (e.g. structured reasoning from a proxy) is ignored instead of failing the whole response. But a non-string, non-null `reasoning_content` now *fails* the decode: before this PR it was an ignored unknown field, and `normalizeReasoning` treats any non-null value as present and hands it to the derived `Option[String]` decoder. The same defensive argument applies to both spellings — consider having `normalizeReasoning` also drop a non-string `reasoning_content` (falling through to the `reasoning` fallback), so a weird provider shape degrades to `None` instead of turning a previously-decodable response into an error.

2. **4× duplicated wrapping boilerplate** — the `{ val derived = ...; Codec.from(derived.prepare(_.withFocus(normalizeReasoning)), derived) }` block appears twice in each of the scala-2 and scala-3 codec files. A small helper next to `normalizeReasoning` in `OpenAIManualCodecs` (shared source) would collapse each site to one call:
   ```scala
   def withReasoningNormalization[A](c: Codec[A]): Codec[A] =
     Codec.from(c.prepare(_.withFocus(normalizeReasoning)), c)
   ```

3. **No user-facing docs** — this is a user-visible feature (and the reason one would point the client at DeepSeek/OpenRouter/Groq in the first place), but nothing in `docs/` mentions `reasoningContent`. A short note in `docs/openai/compatible-apis.md` (which already names these providers) showing that reasoning arrives in `message.reasoningContent` / accumulated `delta.reasoningContent` would make the feature discoverable. Relatedly, the PR description is empty — worth a summary for the changelog.

4. **Binary compatibility** (same note as #551) — new fields on the `Message`/`Delta` case classes break binary compat with 0.5.x (`copy`/`apply`/`unapply` signatures). No MiMa in the build and 0.x versioning, so presumably fine — flagging for release notes.

Also checked: the agent loop only reads `content`/`toolCalls` from the response message and builds request messages from its own types, so the new `reasoning_content` encoder output never leaks into requests sent back to providers.

PR #560, opened 2026-07-31
Feature/timeout retries

### LINKED ISSUE #553: Wire timeout and maxRetries from client config into requests
`AIClientConfig` exposes `timeout` and `maxRetries`, but neither is applied when building requests - the settings are silently ignored across providers. Wire them into request construction (sttp readTimeout / retry wrapper or backend-level policy), add unit tests asserting the request carries the configured timeout, and document the retry semantics (which failures are retried, backoff strategy). This is a prerequisite for any further resilience work in the agent loop (see per-step retry policies).

flsh86:
# Wire timeout and maxRetries from client config into requests

`AIClientConfig` has always exposed `timeout` and `maxRetries`, but neither was applied anywhere — the settings were silently ignored across all providers. This PR makes both real.

## What changed

**Timeout** — every request built by the clients now carries `.readTimeout(config.timeout)`, wired at the single request-construction choke point per provider (`claudeAuthRequest`, `geminiAuthRequest`, `openAIAuthRequest`), so it covers all endpoints including streaming. Semantics are backend-dependent: on the default Java HttpClient backends it is time-to-response-headers, so long SSE streams are not cut off mid-stream. `Duration.Inf` disables the timeout.

**Default timeout raised 60s → 10 minutes** (`DefaultTimeout` on all three configs), matching the official OpenAI/Anthropic SDK defaults. The documented 60s was never honored; making it real would have broken long non-streaming generations.

**Retries** — new shared `sttp.ai.core.http.SyncRetries`, used by the three sync clients (`OpenAISyncClient`, `ClaudeSyncClient`, `GeminiSyncClient`), which own `send`:
- Retried: connection errors (request never reached the server), HTTP 408/429/5xx
- Never retried: read timeouts/read errors (possible double-billing) and other 4xx
- `maxRetries = n` → `n + 1` attempts total; `0` disables
- Exponential backoff 500ms → 1s → 2s → 4s, capped at 8s; delta-seconds `Retry-After` honored (clamped to [0, 30s])
- After exhaustion the last error surfaces unchanged; no new exception types

The async clients return raw sttp `Request`s, so `maxRetries` cannot apply there — documented as bring-your-own-retry (cats-retry, ZIO `Schedule`, …). Same for the agent-loop backends, which send through the caller-provided `Backend`. New docs page: `docs/other/timeouts-and-retries.md`.

## ⚠️ Compatibility

- **Binary-compatibility break (0.5.x)**: `OpenAI` gained a 5th constructor parameter (`timeout: Duration = OpenAIConfig.DefaultTimeout`). Source-compatible; binary-incompatible with 0.5.x artifacts.
- **Behavior changes on upgrade**:
  - Requests now carry a real timeout (default 10 minutes; previously sttp's own 1-minute `basicRequest` default applied on backends honoring `readTimeout`, regardless of the config value).
  - Sync clients now retry transient failures (connect errors, 408/429/5xx) up to `maxRetries` (default 3) times with exponential backoff.
  - Config default `timeout` changed from a never-honored `60.seconds` to `10.minutes`.

## Testing

- `SyncRetriesSpec` (core): 14 cases — retry classification, attempt counting, backoff sequence, caps, ive/overflow clamps), connect-exception retry, read-exception passthrough, interrupt restoration.
- Per-provider timeout specs assert `request.options.readTimeout` on regular and streaming builders, thDuration.Inf` passthrough.
- Per-provider sync-client retry-wiring tests (429 → success), including through OpenAI's structured-ouased overloads.
- Full JVM suite green on Scala 2.13.18 and 3.3.8. (Scala Native rows don't link on the dev machine due — environmental; CI covers them.)
- **Live-validated against the Anthropic API**: all 15 `ClaudeIntegrationSpec` tests pass; a 50ms timeoproving `readTimeout` reaches the backend) while the default succeeds; retry backoff (500ms + 1s)observed on the real send path against a refused connection.


adamw:
## Review

**Overview:** makes the long-ignored `timeout` and `maxRetries` config settings real — `readTimeout` wired at the single request-construction choke point per provider (so all endpoints including streaming builders are covered), and a new shared `SyncRetries` used by the three sync clients. The retry classification (connect errors + 408/429/5xx retried; read errors and other 4xx not, with the double-billing rationale) is sound, and the choke-point placement means nothing was missed — I verified every `.send(backend)` site in the three sync clients now goes through `SyncRetries`, and that no endpoint sends a one-shot `InputStream` request body (uploads are `File`-based), so re-sending is safe.

Verified locally on this branch: `SyncRetriesSpec` (14 tests), all per-provider timeout specs, and the sync-client suites pass on both Scala 2.13 and 3; `scalafmtCheckAll` is clean. The retry loop itself is correct: `maxRetries − attemptNo` gives exactly n+1 attempts, the backoff shift is overflow-guarded, and tail recursion holds.

### Findings

1. **409 Conflict is not retried** — the PR body says the policy matches the official SDKs, but both the OpenAI and Anthropic SDKs retry 408, **409**, 429, and 5xx. 409 is what you get on concurrency conflicts that the SDK vendors consider transient. Either add `StatusCode.Conflict` to `shouldRetry` or drop the "matches the official SDKs" framing for retries (it's accurate for the timeout default).

2. **Clamping `Retry-After` retries earlier than the server instructed** — when a 429 carries `Retry-After: 120`, clamping to 30s means retrying 90 seconds before the server said capacity would return, all but guaranteeing the attempt is burned on another 429. The official SDKs treat an out-of-window `Retry-After` differently (OpenAI's SDK uses it only when it's within ~0–60s, otherwise falls back to its own backoff). Consider: if `Retry-After` exceeds the cap, either fall back to exponential backoff or don't retry at all, rather than retrying early against an explicit server instruction.

3. **`DefaultTimeout`/`DefaultMaxRetries` duplicated three times** — identical `10.minutes`/`3` triples in `OpenAIConfig`, `ClaudeConfig`, `GeminiConfig`. The semantics documentation is already centralized on `AIClientConfig`; the default values could live in its companion (`AIClientConfig.DefaultTimeout`) with the per-provider vals delegating or removed. As-is, a future change has three places to drift.

4. **`customizeRequest` runs once per call, not per attempt** — `sendOrThrow` applies the customization and then `SyncRetries` re-sends that fixed request. Fine for headers, but a customization that computes per-request state (request signing with a timestamp, an idempotency key) won't be refreshed across retries. Worth one sentence in the `customizeRequest` scaladoc so the interaction is explicit.

5. **PR description drift** — it claims an "interrupt restoration" test in `SyncRetriesSpec`, but no such test exists in the diff (the handling itself is in the code, untested). Also several sentences in the Testing section are visibly garbled/truncated ("thDuration.Inf", "ive/overflow clamps", "50ms timeoproving"). Worth fixing the description, and ideally adding the missing interrupt test since the description promises it.

Smaller notes:
- The backend-dependent `readTimeout` semantics (time-to-headers vs. between-bytes) are documented honestly in three places — good; the `Duration.Inf` escape hatch is tested per provider.
- The non-config `OpenAISyncClient.apply` overloads can't set `timeout`/`maxRetries`, so they get the new defaults (10 min, 3 retries). That's the documented behavior change and the config path covers customization, so no action needed — just confirming it was considered.
- Binary-compatibility break on the `OpenAI` constructor is correctly called out in the PR body; consistent with #551/#552, no MiMa gating exists.

adamw:
## Requested change: implement retries as an sttp backend wrapper

Let's keep the retries, but change the shape. `SyncRetries` applied at each sync-client call site means retries only exist for the sync clients, and every new endpoint has to remember to route through it. Instead, please implement retries as an **sttp backend wrapper**, so they are optional, composable, and work across the board — sync clients, async clients, and the agent loops alike.

Concretely:

- Add a `RetryingBackend` to `core`, following sttp's wrapper pattern (`DelegateBackend`), generic in the effect and **parametrized with a way to sleep** — `sleep: FiniteDuration => F[Unit]`. That solves the reason retries were sync-only in the first place (sttp's `MonadError` has no timer): the sync instance passes `Thread.sleep`, and async users pass their effect's timer (`IO.sleep`, `ZIO.sleep`, pekko `after`, Ox `sleep`) — each a one-liner we can document or ship as thin helpers.
- Use sttp's built-in retry machinery rather than a hand-rolled loop where possible: `RetryWhen.Default` as the base predicate for "is this request safe to retry at all" (idempotent method, re-sendable body — this matters once arbitrary user requests can flow through the wrapper, not just the library's own), extended with this PR's response classification (408/429/5xx, `Retry-After`) — incorporating the earlier findings (409, and falling back to backoff instead of clamping an out-of-window `Retry-After`).
- The sync clients keep honoring `config.maxRetries` simply by wrapping the backend they construct/receive (`maxRetries = 0` → no wrapper), so the config-driven behavior of this PR is preserved and `SyncRetries` disappears.
- The async story in `docs/other/timeouts-and-retries.md` then changes from "bring your own retry library" to: wrap your backend with `RetryingBackend` and your effect's sleep — with a short cats-effect or ZIO example.

Rough shape (adjust to sttp4's actual wrapper API):

```scala
class RetryingBackend[F[_], P](
    delegate: GenericBackend[F, P],
    sleep: FiniteDuration => F[Unit],
    maxRetries: Int
) extends DelegateBackend(delegate) {
  override def send[T](request: GenericRequest[T, P with Effect[F]]): F[Response[T]] = ...
}
```

One caveat to handle explicitly: responses whose bodies are streams (SSE/`InputStream`) can't be retried once the body handler has run — the wrapper should only retry when the response body is safely replayable (for the error statuses we retry, the body is small and fully read, so in practice this falls out of the `RetryWhen`-style guard, but it deserves a test).

The retry policy itself and the test coverage from this PR are good — this is about relocating the mechanism, not redesigning the policy.

flsh86:
# Wire timeout and maxRetries from client config into requests

`AIClientConfig` has always exposed `timeout` and `maxRetries`, but neither was applied anywhere — the settings were silently ignored across all providers. This PR makes both real.

## What changed

**Timeout** — every request built by the clients now carries `.readTimeout(config.timeout)`, wired at the single request-construction choke point per provider (`claudeAuthRequest`, `geminiAuthRequest`, `openAIAuthRequest`), so it covers all endpoints including streaming. Semantics are backend-dependent: on the default Java HttpClient backends it is time-to-response-headers, so long SSE streams are not cut off mid-stream. `Duration.Inf` disables the timeout.

**Default timeout raised 60s → 10 minutes** (`AIClientConfig.DefaultTimeout`, delegated to by the three config companions), matching the official OpenAI/Anthropic SDK defaults. The documented 60s was never honored; making it real would have broken long non-streaming generations.

**Retries** — implemented as an sttp **backend wrapper**, `sttp.ai.core.http.RetryingBackend` (in the style of `FollowRedirectsBackend`), generic in the effect with caller-supplied sleep. The sync clients (`OpenAISyncClient`, `ClaudeSyncClient`, `GeminiSyncClient`) wrap their backend automatically to honor `maxRetries`; async clients and agent loops opt in by wrapping the backend they use (`RetryingBackend(backend, maxRetries, IO.sleep)` etc.):
- Retried: connection errors (request never reached the server), HTTP 408/409/429/5xx — the set the official OpenAI/Anthropic SDKs retry
- Never retried: read timeouts/read errors (possible double-billing), other 4xx, and requests whose body cannot be re-sent (streaming bodies)
- `maxRetries = n` → `n + 1` attempts total; `0` disables
- Exponential backoff 500ms → 1s → 2s → 4s, capped at 8s; a delta-seconds `Retry-After` within [0, 30s] is honored, outside that window the exponential backoff applies instead of retrying earlier than the server instructed
- A retried response's body is closed first when `AutoCloseable`; after exhaustion the last error surfaces unchanged; no new exception types

`customizeRequest` (OpenAI sync client) runs once per call: the same customized request is re-sent on every attempt — documented in its scaladoc. New docs page: `docs/other/timeouts-and-retries.md`.

## ⚠️ Compatibility

- **Binary-compatibility break (0.5.x)**: `OpenAI` gained a 5th constructor parameter (`timeout: Duration = OpenAIConfig.DefaultTimeout`). Source-compatible; binary-incompatible with 0.5.x artifacts.
- **Behavior changes on upgrade**:
  - Requests now carry a real timeout (default 10 minutes; previously sttp's own 1-minute `basicRequest` default applied on backends honoring `readTimeout`, regardless of the config value).
  - Sync clients now retry transient failures (connect errors, 408/409/429/5xx) up to `maxRetries` (default 3) times with exponential backoff.
  - Config default `timeout` changed from a never-honored `60.seconds` to `10.minutes`.

## Testing

- `RetryingBackendSpec` (core, 18 cases): retry classification (incl. 409), attempt counting, backoff sequence and caps, `Retry-After` window handling (honored within 0–30s; date-form/negative/out-of-window fall back to backoff), connect-exception retry, read-exception passthrough, non-resendable request bodies never retried, `AutoCloseable` response body closed before retry, interrupt-flag restoration on interrupted sleep, and retries through an asynchronous (`Future`) backend via the effect's sleep.
- Per-provider timeout specs assert `request.options.readTimeout` on regular and streaming builders, the 10-minute default, and `Duration.Inf` passthrough.
- Per-provider sync-client retry-wiring tests (429 → success), including through OpenAI's structured-output path and the config-based overloads.
- Full JVM suite green on Scala 2.13.18 and 3.3.8. (Scala Native rows don't link on the dev machine due to a local x86_64 libidn2 — environmental; CI covers them.)
- Live-validated against the Anthropic API: all 15 `ClaudeIntegrationSpec` tests pass; a 50ms timeout aborts a real request while the default succeeds; retry backoff (500ms + 1s) observed on the real send path against a refused connection.

PR #568, opened 2026-08-04
Feature/agent testkit

### LINKED ISSUE #555: AgenticFunction - module with a scripted agent backend
Users currently have no way to test agent code without hitting a paid API. Add a published testkit module providing: (1) a ScriptedAgentBackend that returns a queued sequence of assistant messages/tool calls and records every request for assertions; (2) matchers for asserting on prompts, tool schemas sent, and tool-call arguments; (3) optional ScalaCheck integration deriving Arbitrary[T] for tool input types so generators and schemas share one source of truth. Embabel treats its testing library as a headline feature — this is the cheapest credibility win for sttp-ai’s agent story.

flsh86:
Adds a published `testkit` module so users can test agent code without calling a paid LLM API — the agent-loop equivalent of sttp-client's backend stubs.

## What's included

- **`ScriptedAgent[F]`** — scripts the model's responses (one per round-trip) and produces a standard `AgentBuilder`, so swapping a real agent for a scripted one is a one-line change (`OpenAIAgent.builder(...)` → `script.builder`). The real agent loop runs: user tools actually execute.
- **`ScriptedResponse`** — readable response constructors: `text`, `toolCall`, `toolCalls`, `textWithToolCalls`, `maxTokens` (auto-generated `call_N` ids).
- **`ScriptedAgentBackend[F]`** — the underlying `AgentBackend`: dequeues scripted responses inside the effect (`monad.suspend`, correct for IO/ZIO), records every request, and fails fast with `ScriptExhaustedException` when the script runs out.
- **`RecordedInteractions`** — framework-agnostic query API: full histories, `includeTools` per call, offered tool schemas (`rawJsonSchema`), prompts, tool results, system prompt, configured response schema.
- **`AgentMatchers`** — scalatest matchers: `haveReceivedPrompt`, `haveOfferedTool`, `haveOfferedToolWithSchema`, `haveCalledTool`, `haveCalledToolWith` (JSON-structural argument comparison), `haveFinishedWith`.
- **Docs**: new `docs/agents/testing.md` page (mdoc-compiled worked example), index + README mentions.

## Module notes

- Cross-built like `core`: Scala 2.13 + 3 on JVM, Scala 3 on Native; published as `com.softwaremill.sttp.ai::testkit` (add in `Test` scope).
- scalatest is intentionally a compile-scope dependency (the module ships matcher implementations); non-scalatest users can use the plain `RecordedInteractions` query API.
- No changes to `core` — the existing `AgentBackend`/`AgentBuilder` seam is used as-is.
- No `examples/` entry yet: CI compiles examples against published artifacts, so a runnable example lands after the first release containing `testkit`.
- Follow-up (out of scope here): optional ScalaCheck integration deriving `Arbitrary[T]` for tool input types from the tapir schema.

## Testing

33 unit tests per Scala version (66 total), driving the real `Agent` loop end-to-end — tool round-trips, last-iteration tool withholding, exhaustion, `runAs` decoding, response-schema recording, matcher pass/failure messages. `sbt compileDocumentation` verifies the docs snippet compiles. No integration tests — nothing touches a network, which is the point.


docs/agents/testing.md
```diff
@@ -0,0 +1,86 @@
+# Testing agents
+
+The `testkit` module lets you test agent code without hitting a paid LLM API: you script the model's responses, run the real agent loop against them, and assert on what was sent to the "model" and what the agent did.
```
adamw: Maybe it would be useful to describe the use-case: providing a stub agent which is scripted to invoke certain functions (tools), and which is scripted to provide a given final response. So this can be used for integration testing where an agent is a parameter of a larger process?

adamw:
## Overview

The design is sound. Reusing `AgentBuilder` so the swap is one line (`OpenAIAgent.builder(...)` → `script.builder`) is the right call, and running the real `Agent` loop means tests actually exercise tool dispatch, argument decoding, and last-iteration tool withholding.

I checked that `monad.suspend` is used correctly, and that the `[Iteration $current of $max]` rendering in `RecordedInteractions.userPrompts` matches what OpenAI, Claude and Gemini all really send — so that accessor is faithful.

## Main concerns

**1. scalatest in compile scope (`project/Dependencies.scala:24`)**

This is the one that's expensive to change after publishing. Everyone depending on `testkit` gets scalatest 3.2.20 on their compile classpath, including people who only want `RecordedInteractions`, and it can clash with their own scalatest version.

`Provided` scope fixes it in one line: the matchers still compile, and users bring their own scalatest — which every scalatest user already has in `Test`. Splitting into `testkit` / `testkit-scalatest` also works, but it's more machinery than the problem needs.

**2. The artifact is called `testkit`, but it only tests agents**

sbt-projectmatrix will publish `com.softwaremill.sttp.ai:testkit_3`. Everything in it is agent-loop specific (package `sttp.ai.core.agent.testing`). If a client-level testkit ever shows up, the obvious name is already taken. `agent-testkit` costs nothing right now.

**3. `offeredTools` only looks at the first request (`RecordedInteractions.scala:339`)**

```scala
final def offeredTools: Seq[OfferedTool] = requests.headOption.map(_.toolsOffered).getOrElse(Seq.empty)
```

With `maxIterations(1)` the loop sets `includeTools = false` on the only call, so this is empty and `haveOfferedTool("calculator")` fails with `offered tools: none`. That reads like "you forgot to register the tool", when the tool was registered and deliberately withheld. The scaladoc mentions this, but the failure message is where people will actually hit it.

`requests.flatMap(_.toolsOffered).distinctBy(_.name)` matches what the matcher promises ("was this tool offered") and removes the trap. If you keep head-only, at least make the matcher say when no call offered any tools.

Smaller, related: the accessors mix `head` (`initialPrompt`, `offeredTools`, `systemPromptSent`, `responseSchemaSent`) and `last` (`userPrompts`, `toolResultsSent`). Each one is justified and documented, but it's a rule users have to remember.

## Test gaps

- **No effect type other than `Identity` is tested.** The description calls out `monad.suspend` as "correct for IO/ZIO", but nothing checks it. `sttp.monad.TryMonad` needs no new dependency and would let you assert the script isn't consumed until the effect runs.
- **`ScriptedResponse.maxTokens` never drives the loop.** Only the constructor is tested; nothing scripts it and asserts `FinishReason.TokenLimit`. Same for `haveFinishedWith(FinishReason.TokenLimit)`, which is only tested in its failing branch.
- **`textWithToolCalls` never goes through the loop** — constructor test only.
- **No test for a scripted call to a tool that isn't registered.** `Agent` returns `"Tool not found: X"` there; that's a realistic thing to script and assert on.
- `AgentMatchersSpec.scala:558` — the test is called "fail showing both schemas when they differ" but only asserts `include("calculator")`. It never checks that either schema shows up.

## Smaller points

- `RecordedRequest.responseSchema` (`RecordedInteractions.scala:314`) has a `= None` default, but it's built in exactly one place with all arguments. The default only makes a future omission silent — drop it.
- `ScriptedAgentBackend` isn't `final`, while `ScriptedAgent` is. Direct instantiation is intended, subclassing probably isn't.
- Calling `.build` twice on a kept builder quietly creates a second backend, replays the script from the start and doubles `script.requests`. It's in the scaladoc and there's a test for it, but it's an easy mistake with a confusing symptom. Worth a sentence in `docs/agents/testing.md`, which doesn't mention it today.
- The docs example is `mdoc:compile-only`, so the main example of a testing library never actually runs. It's fully offline so running it would be real verification, though mdoc running a scalatest spec isn't natural. Fine to leave.

## Checked, no issues

- Loop integration — `includeTools`, iteration markers, `runAs` decoding and exhaustion all line up with `Agent.scala`.
- Variance — `Matcher[AgentResult[_]]` works against `AgentResult[String]` because scalatest's `Matcher` is contravariant; the tests confirm it.
- Thread safety — the `synchronized` blocks are consistent, and the documented caveat about cross-build ordering is accurate.
- Security — nothing touches the network, filesystem or credentials.
- Build wiring — `allAgregates`, the docs `dependsOn` and the JVM 2.13+3 / Native 3 matrix all match how `core` is set up.

## Recommendation

Good to merge after (1) and (3). Both change the published API, so they're much cheaper to fix before the first release than after. The test gaps and (2) are worth doing but wouldn't block.

PR #570, opened 2026-08-05
Feature/capability tagged models

### LINKED ISSUE #558: Capability-tagged model constants and per-step model selection
Tag generated model constants with capability marker traits (Vision, ToolCalling, StructuredOutput, Reasoning) maintained by the model-update scripts, so APIs can require M <: Model & Vision and reject invalid model/task pairings at compile time. Then allow the agent loop to use different models per phase (e.g. a cheap model for tool-selection iterations, a stronger model for the final synthesis) — Embabel’s “LLM mixing” — starting with a simple modelForIteration: (IterationInfo => M) config hook

flsh86:
Capability-tagged model constants (all three providers)

- New marker traits in core — sttp.ai.core.model.Capability.{Vision, ToolCalling, StructuredOutput, Reasoning} — plus a Capability.All shorthand extending all four, and Supports[M, C] implicit evidence with a readable @implicitNotFound message. Supports is sealed by design: capabilities are declared by mixing traits into the model type, and custom model classes are the intended escape hatch.
- OpenAI ChatCompletionModel, ClaudeModel, and GeminiModel constants mix in the capabilities they support (e.g. GPT4o is Vision with ToolCalling with StructuredOutput; O1Mini is Reasoning only; ClaudeSonnet5 is Capability.All). OpenAI tags are maintained by the model-update script from a curated mapping; Claude/Gemini are hand-maintained.
- Custom/unknown model classes (CustomChatCompletionModel, new CustomClaudeModel, GeminiModel.CustomModel) claim all capabilities — using a raw model-name string asserts your model supports what you use it for, so OpenAI-compatible providers (Ollama, Groq, OpenRouter) keep working unchanged.

Compile-time enforcement in the agent APIs

- AgentBuilder gains a phantom model type: AgentBuilder[F, M <: AIModel]. .tools(...)/.addTool require ToolCalling on M; .responseSchema/.deriveResponseSchema require StructuredOutput. An agent with no tools accepts any model.
- Typed provider entry points: ClaudeAgent.builder[F](client, ClaudeModel.ClaudeSonnet5) infers the model's capabilities, so invalid pairings — .tools on o1-mini, .deriveResponseSchema on claude-3-haiku — fail to compile. All modelName: String call shapes keep compiling and skip capability checking.

Per-iteration model selection ("LLM mixing")

- The loop passes IterationInfo(iteration, maxIterations) (1-based; isLastIteration flags the forced-final iteration where tools are withheld) to AgentBackend.sendRequest.
- Provider builders accept modelForIteration: IterationInfo => M — e.g. a cheap model for tool iterations and a stronger one for the forced-final synthesis. The inferred M covers every model the hook can return, so capability checks require what all of them share. Same provider only; cross-provider mixing is out of scope.

Tooling, testkit & docs

- update_code_with_new_models.scala maintains the mixin clauses from a hand-curated, first-match-wins capabilities section in model_update_config.yaml (never scraped). It re-renders every constant in the managed block on each run (config is the source of truth), renders Capability.All for full-capability models, tolerates scalafmt-wrapped definitions, and validates all capability names eagerly at config load. The config's stale sttp/openai/... paths were fixed to sttp/ai/openai/....
- agent-testkit: ScriptedModel stand-in claims all capabilities so scripted tests never fight evidence; RecordedRequest records the per-request IterationInfo.
- Docs: new "Model capabilities" and "Per-iteration model selection" sections in docs/agents/configuration.md as mdoc:compile-only snippets (pinned against API drift); updated AgentBackend signature in custom-backends.md; quickstart uses a typed constant. examples/ untouched (compiles against released artifacts).

⚠️ Breaking changes (0.x, binary-incompatible)

- AgentBuilder[F] → AgentBuilder[F, M <: AIModel] (type annotations on builders need updating).
- AgentBackend.sendRequest gains a fourth parameter, iterationInfo: IterationInfo (affects custom backe
- ClaudeModel changed from sealed trait to sealed abstract class ClaudeModel(val value: String) extendsult and ClaudeModel.WithStructuredOutput are removed — use the Capability.StructuredOutput tag instead
(modelSupportsStructuredOutput keeps its exact semantics, unknown ⇒ supported).
- RecordedRequest (agent-testkit) gains an iterationInfo field (constructor arity change).
- Provider builder[F] methods became partially-applied applicators (builder[F] returns a class with appd call shapes (builder[IO](client, "name"), synchronous(...)) remain source-compatible, but binary
compatibility is broken and exotic usages (eta-expansion of builder) change.
- ScriptedAgent.builder (agent-testkit) now returns AgentBuilder[F, ScriptedModel.type].

Wire formats are unchanged — the model field sent to each API is still the same string.

Testing

- Compile-time behavior tested with assertDoesNotCompile/assertCompiles pairs (every negative has a posi, and claude, including Capability.All evidence resolution.
- Per-iteration model selection verified at the wire level for all three providers (stubbed HTTP, assersent per request); testkit's IterationInfo recording asserted.
- Claude structured-output regression contract preserved verbatim (incl. unknown-model ⇒ supported).
- Script verified byte-idempotent (--apply + sbt scalafmtAll twice → identical file), and a typo'd caparule fails loudly at config load.
- Full JVM suite green on Scala 2.13 and 3 (1188+ tests, 0 failures); compileDocumentation green.

### DIFF
--- a/docs/agents/configuration.md
+++ b/docs/agents/configuration.md
@@ -27,7 +27,8 @@ properties nullable); when `false`, tools are registered as non-strict with thei
 ## Model capabilities
 
 Model constants are tagged with capability marker traits: `Vision`, `ToolCalling`, `StructuredOutput`, `Reasoning`
-(in `sttp.ai.core.model.Capability`). Agent builder methods that need a capability require it at compile time:
+(in `sttp.ai.core.model.Capability`; models supporting everything mix in the `Capability.All` shorthand, which
+extends all four). Agent builder methods that need a capability require it at compile time:
 `.tools`/`.addTool` need `ToolCalling`, `.responseSchema`/`.deriveResponseSchema` need `StructuredOutput`.
 
 ```scala mdoc:compile-only
--- a/docs/agents/configuration.md
+++ b/docs/agents/configuration.md
@@ -30,8 +30,14 @@ Model constants are tagged with capability marker traits: `Vision`, `ToolCalling
 (in `sttp.ai.core.model.Capability`). Agent builder methods that need a capability require it at compile time:
 `.tools`/`.addTool` need `ToolCalling`, `.responseSchema`/`.deriveResponseSchema` need `StructuredOutput`.
 
-```scala
+```scala mdoc:compile-only
+import sttp.ai.core.agent.AgentTool
+import sttp.ai.openai.OpenAI
+import sttp.ai.openai.agent.OpenAIAgent
 import sttp.ai.openai.requests.completions.chat.ChatRequestBody.ChatCompletionModel
+import sttp.shared.Identity
+
+def myTool: AgentTool[Identity, ?] = ???
 
 val agent = OpenAIAgent
   .synchronous(OpenAI.fromEnv, ChatCompletionModel.GPT4o) // GPT4o mixes in ToolCalling
@@ -44,7 +50,14 @@ val agent = OpenAIAgent
 String model names keep working and skip capability checking — they wrap into the provider's custom model class,
 which claims all capabilities (you assert your model supports what you use it for):
 
-```scala
+```scala mdoc:compile-only
+import sttp.ai.core.agent.AgentTool
+import sttp.ai.openai.OpenAI
+import sttp.ai.openai.agent.OpenAIAgent
+import sttp.shared.Identity
+
+def myTool: AgentTool[Identity, ?] = ???
+
 val ollamaAgent = OpenAIAgent.synchronous(OpenAI.fromEnv, "llama3-70b").tools(myTool).build
 ```
 
@@ -53,8 +66,14 @@ val ollamaAgent = OpenAIAgent.synchronous(OpenAI.fromEnv, "llama3-70b").tools(my
 Agents can use a different model per loop iteration — e.g. a cheap model while the agent calls tools, and a stronger
 model for the forced-final iteration (where tools are withheld and the model must answer):
 
-```scala
-import sttp.ai.core.agent.IterationInfo
+```scala mdoc:compile-only
+import sttp.ai.core.agent.{AgentTool, IterationInfo}
+import sttp.ai.openai.OpenAI
+import sttp.ai.openai.agent.OpenAIAgent
+import sttp.ai.openai.requests.completions.chat.ChatRequestBody.ChatCompletionModel
+import sttp.shared.Identity
+
+def myTool: AgentTool[Identity, ?] = ???
 
 val mixedAgent = OpenAIAgent
   .synchronous(OpenAI.fromEnv, (info: IterationInfo) => if info.isLastIteration then ChatCompletionModel.GPT5 else ChatCompletionModel.GPT4oMini)
@@ -65,7 +84,10 @@ val mixedAgent = OpenAIAgent
 The inferred model type is the least upper bound of every model the function can return, so capability checks require
 the capabilities **all** of them share. If inference produces an unwieldy type, ascribe the function explicitly:
 
-```scala
+```scala mdoc:compile-only
+import sttp.ai.core.agent.IterationInfo
+import sttp.ai.openai.requests.completions.chat.ChatRequestBody.ChatCompletionModel
+
 val pick: IterationInfo => ChatCompletionModel & sttp.ai.core.model.Capability.ToolCalling =
   info => if info.isLastIteration then ChatCompletionModel.GPT5 else ChatCompletionModel.GPT4oMini
 ```
--- a/docs/agents/configuration.md
+++ b/docs/agents/configuration.md
@@ -24,6 +24,55 @@ The OpenAI factories additionally accept a `strictTools` flag (default `true`): 
 normalized for OpenAI's strict function calling (`additionalProperties: false`, all properties required, optional
 properties nullable); when `false`, tools are registered as non-strict with their original schemas.
 
+## Model capabilities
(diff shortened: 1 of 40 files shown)


docs/agents/configuration.md
```diff
@@ -24,6 +24,78 @@ The OpenAI factories additionally accept a `strictTools` flag (default `true`):
...
+```
+
+The inferred model type is the least upper bound of every model the function can return, so capability checks require
+the capabilities **all** of them share. If inference produces an unwieldy type, ascribe the function explicitly:
+
+```scala mdoc:compile-only
+import sttp.ai.core.agent.IterationInfo
+import sttp.ai.openai.requests.completions.chat.ChatRequestBody.ChatCompletionModel
+
+val pick: IterationInfo => ChatCompletionModel & sttp.ai.core.model.Capability.ToolCalling =
```
adamw: maybe drop the fully qualified name & add an import?

adamw:
Do we have docs on how to define custom models with capabilities? Not all will be always present in our constants

adamw:
## Review

I checked the mechanical parts first: the wire format is unchanged (`chatCompletionModelCodec` encodes `.value`), `IterationInfo(iteration + 1, maxIterations).isLastIteration` matches the loop's own `isLastIteration` (`iteration == maxIterations - 1`), `modelSupportsStructuredOutput` keeps its exact semantics including unknown ⇒ true, and every `AgentBackend` implementation in the repo is updated. That all looks right.

Below are the things I'd want changed or at least discussed.

### Capability data — two concrete errors

The tags are hand-curated, and wrong tags are the main failure mode of this design. Two I'm confident about:

- **`gpt-4-0314` and `gpt-4-32k-0314` are tagged `ToolCalling`** (via `defaults`). Function calling arrived with the `-0613` snapshots; the `0314` ones don't support it. The config explicitly zeroes out `gpt-3.5-turbo-0301` for exactly this reason but misses the gpt-4 equivalents — add a `pattern: "gpt-4*-0314"` rule with `capabilities: []`.
- **`gpt-4o-2024-05-13` is tagged `StructuredOutput`** via the `gpt-4o*` catch-all. The `json_schema` response format needs `gpt-4o-2024-08-06` or later. It needs its own rule above the catch-all.

It would also help to note in the yaml which date/source the table was curated from, so the next person knows how stale it is.

### No escape hatch for a wrong tag

`Supports` is sealed and `Supports.instance` is private, so when a tag is missing or wrong (see above), the only option is to drop to the `String` overload — which turns off *all* capability checking, not just the one that's wrong. Since the tags will drift as providers change, I'd add:

```scala
def assume[M, C]: Supports[M, C] = instance.asInstanceOf[Supports[M, C]]
```

Then a user can write `given Supports[GPT4o20240513.type, StructuredOutput] = Supports.assume` and keep the rest of the checking. Cheap, and it turns a hard block into an explicit opt-out.

### API notes

- **Widening silently disables the checks.** `val m: ChatCompletionModel = ChatCompletionModel.GPT4o` and then `.tools(...)` will not compile, because `M` infers to the base type. Same for a model held in a config case class. The docs cover ascription for the function case only — one sentence for the plain value case would help.
- **`Vision` and `Reasoning` aren't used by any API.** That's fine as documentation, but it means each new model needs four correct answers where only two do any work.

### Model-update script

- **Endpoint keys under `capabilities` aren't validated against `endpoints`.** `validateCapabilityNames` catches a typo'd capability name, but a typo'd endpoint key silently does nothing. Two lines in the same loop.
- **`isCaseObjectStart` and `parseCaseObjects` disagree.** The first skips blank lines when looking for the `extends` continuation (`lines.drop(index+1).find(_.trim.nonEmpty)`); the second requires it at exactly `i + 1`. Scalafmt won't put a blank line there, so it's latent, but both should use one helper.
- **The parser has no tests.** It went from a one-line regex to a hand-rolled mutable index loop handling two shapes. Idempotency was checked by hand; a small unit test over `parseCaseObjects` with a wrapped and an unwrapped definition would lock that in cheaply.

### Smaller things

- `core/.../model/AIModel.scala` holds `AIModel`, `Capability` and `Supports`. The interesting parts aren't in the file name, and the test is `CapabilitySpec.scala`. Split it, or rename to `Capability.scala`.
- `IterationInfo` is `case class`, not `final case class` — `RecordedRequest` and `OfferedTool` are both final.
- `RecordedRequest.iterationInfo` has no default, so every testkit user building one by hand breaks. A default would soften that at no cost.
- Three near-identical `BuilderPartiallyApplied` classes plus 6-8 `synchronous` overloads each. Acceptable given the provider-specific types, but OpenAI is missing `apply(apiKey: String, model: M)` — a typed model only works through the `OpenAI` entry point, not the apiKey one.
- `OpenAIAgentCapabilitySpecFixtures.openAI` has no type annotation on a public member.
- No migration or changelog note in `docs/` for a breaking change of this size.

### Tests

The `assertDoesNotCompile` / `assertCompiles` pairing is the right approach, per-iteration selection is checked at the wire level for all three providers, and `AgentSpec` covers the 1-based numbering and the forced-final flag. Gaps: no negative test for `.responseSchema` (only `deriveResponseSchema`), and nothing covers the script parser.

### Summary

The mechanics are sound and well tested. The two things I'd want before merge are the `Supports.assume` escape hatch (or something like it) and the two mis-tagged OpenAI models — together they're what makes a wrong tag recoverable instead of a hard block. The endpoint-key validation and a script parser test are cheap follow-ups.

PR #574, opened 2026-08-06
Feature/agent interceptor

flsh86:
## Description

Generalizes the ad-hoc agent hooks (`beforeToolCall`, `afterToolCall`) into a composable `AgentInterceptor[F]` middleware that wraps iterations, LLM calls, and tool executions — with LLM-call visibility and budget enforcement as the first built-in deliverables.

### AgentInterceptor[F]

```scala
trait AgentInterceptor[F[_]] {
  def aroundIteration[A](ctx: IterationContext)(next: => F[A]): F[A] = next
  def aroundLlmCall(ctx: LlmCallContext)(next: => F[AgentResponse]): F[AgentResponse] = next
  def aroundToolCall(ctx: ToolCallContext)(next: => F[ToolCallRecord]): F[ToolCallRecord] = next
  def decide(state: AgentRunState): LoopDecision = LoopDecision.Continue
}
```

- Wrap-style (onion) composition like sttp backend wrappers: `agent.interceptor(a).interceptor(b)` — first added is outermost; for `decide`, the first `FinishNow` wins.
- `next` is by-name so interceptors work under `Identity` as well as lazy effect types (contract: call `next` at most once).
- `decide` is pure and consulted before each iteration; returning `LoopDecision.FinishNow(cause, instruction)` injects the instruction as a user message and withholds tools — mirroring the existing last-iteration behavior — instead of failing abruptly.

### Usage accounting

- `AgentResponse` now carries provider-reported `usage: Option[TokenUsage]` and `model: Option[String]`; all three backends (OpenAI, Claude incl. cache fields, Gemini) populate them.
- New `Tokens` value class and provider-normalized `TokenUsage` (input includes cached; output includes reasoning; cached/reasoning as informational subsets).
- `AgentResult` accumulates `usage: TokenUsage` and a per-call `llmCalls: Seq[LlmCallUsage]` breakdown (model id + usage per call, enabling per-model cost math for mixed-model runs).
- `FinishReason` gains `BudgetExceeded`; `runAs` treats budget-forced answers like max-iterations ones (parse-attempted).

### Built-in interceptors (`sttp.ai.core.agent.interceptor`)

- **`BudgetInterceptor`** — `maxTotalTokens` and/or `maxCost` with a user-supplied `PriceTable` (the library ships no prices). On breach it forces a graceful final answer with `FinishReason.BudgetExceeded`. Setting `maxCost` without a `priceTable` fails fast at construction; calls whose model id is missing from the table contribute zero to the cost check (documented loudly); the token budget needs no table. Budgets are soft by one LLM call (the forced final answer) — documented.
- **`LoggingInterceptor`** — pluggable sink `(LogLevel, String) => F[Unit]`, no logging dependency in core; docs show slf4j/log4cats, ZIO logging, and otel4s bridges.

### Compatibility

- ⚠️ **Binary compatibility break** (0.8.0): `AgentResponse`, `AgentResult`, and `AgentConfig` constructors gained fields. All new fields are defaulted, so **source compatibility is preserved** for typical usage, including external `AgentBackend` implementations. One source-level exception: `FinishReason` (sealed) gained the `BudgetEe with exhaustive matches on `FinishReason` must add a case.
- `beforeToolCall`/`afterToolCall` (config fields) and `hookBeforeToolCall`/`hookAfterToolCall` (builder) are `@deprecated` since 0.8.0 and adapt internally to an interceptor running innermost — behavior unchanged for one release,    removal planned for the next.
                                                                                                                                                                                                                                          ### Docs
                                                                                                                                                                                                                                          New page `docs/agents/interceptors.md` (registered in the toctree): writing interceptors, composition/o and price-table caveats, logging bridges, hook migration. The existing hooks section in`docs/agents/configuration.md` now carries a deprecation admonition.                                                                                                                                                                      
### Testing                                                                                                                                                                                                                               
- New unit suites: `TokenUsageSpec`, `AgentInterceptorComposeSpec`, `AgentInterceptorLoopSpec` (onion ordering, usage accumulation, steering incl. iteration-0 FinishNow, maxIterations precedence, legacy-hook interleaving, interceptor error propagation), `BudgetInterceptorSpec` (incl. multi-model cost math), `LoggingInterceptorSpec`.
- Provider backend specs assert the exact usage mappings against stubbed JSON.
- One cheap assertion added to the shared agent integration base (`usage.totalTokens > 0`) — no new API
- Full suite green on Scala 2.13 and 3 JVM rows; `compileDocumentation` passes.
- **Live-validated against real APIs:** Claude 8/8 (including an end-to-end budget-forced graceful stop), Gemini 7/7, OpenAI 6/7 — the one OpenAI failure (`runAs[T]` expecting `NaturalStop`) reproduces

### DIFF
--- a/core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
@@ -53,8 +53,10 @@ final class AgentBuilder[F[_], M <: AIModel] private (
   ): AgentBuilder[F, M] =
     withConfig(config.copy(responseSchema = Some(ResponseSchema.derived[T](Some(description)))))
 
+  @deprecated("Use interceptor(...) with an AgentInterceptor overriding aroundToolCall", "0.8.0")
   def hookBeforeToolCall(hook: ToolCall => F[Unit]): AgentBuilder[F, M] = withConfig(config.copy(beforeToolCall = Some(hook)))
 
+  @deprecated("Use interceptor(...) with an AgentInterceptor overriding aroundToolCall", "0.8.0")
   def hookAfterToolCall(hook: ToolCallRecord => F[Unit]): AgentBuilder[F, M] = withConfig(config.copy(afterToolCall = Some(hook)))
 
   /** Appends an interceptor. Interceptors wrap stages in list order: the first added is outermost. */
--- a/core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
@@ -53,14 +53,8 @@ final class AgentBuilder[F[_], M <: AIModel] private (
   ): AgentBuilder[F, M] =
     withConfig(config.copy(responseSchema = Some(ResponseSchema.derived[T](Some(description)))))
 
-  @deprecated("Use interceptor(...) with an AgentInterceptor overriding aroundToolCall", "0.8.0")
-  def hookBeforeToolCall(hook: ToolCall => F[Unit]): AgentBuilder[F, M] = withConfig(config.copy(beforeToolCall = Some(hook)))
-
-  @deprecated("Use interceptor(...) with an AgentInterceptor overriding aroundToolCall", "0.8.0")
-  def hookAfterToolCall(hook: ToolCallRecord => F[Unit]): AgentBuilder[F, M] = withConfig(config.copy(afterToolCall = Some(hook)))
-
   /** Appends an interceptor. Interceptors wrap stages in list order: the first added is outermost. */
-  def interceptor(value: AgentInterceptor[F]): AgentBuilder[F, M] =
+  def addInterceptor(value: AgentInterceptor[F]): AgentBuilder[F, M] =
     withConfig(config.copy(interceptors = config.interceptors :+ value))
 
   /** Replaces the interceptor list. */
--- a/core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
@@ -57,6 +57,14 @@ final class AgentBuilder[F[_], M <: AIModel] private (
 
   def hookAfterToolCall(hook: ToolCallRecord => F[Unit]): AgentBuilder[F, M] = withConfig(config.copy(afterToolCall = Some(hook)))
 
+  /** Appends an interceptor. Interceptors wrap stages in list order: the first added is outermost. */
+  def interceptor(value: AgentInterceptor[F]): AgentBuilder[F, M] =
+    withConfig(config.copy(interceptors = config.interceptors :+ value))
+
+  /** Replaces the interceptor list. */
+  def interceptors(values: Seq[AgentInterceptor[F]]): AgentBuilder[F, M] =
+    withConfig(config.copy(interceptors = values))
+
   def build: Agent[F] = Agent(makeBackend(config), config)
 }
 
--- a/core/src/main/scala/sttp/ai/core/agent/AgentInterceptor.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/AgentInterceptor.scala
@@ -87,7 +87,8 @@ object LoopDecision {
 
   /** Force a graceful final answer: `instruction` is injected as a user message, tools are withheld (the existing last-iteration
     * mechanism), and the run finishes with `finishReason = cause`. If the loop is simultaneously at its forced last iteration,
-    * `MaxIterations` takes precedence as the reported reason.
+    * `MaxIterations` takes precedence as the reported reason. If the forced-final LLM response itself stops with `StopReason.MaxTokens`,
+    * that check precedes the `FinishNow` cause and the run reports `FinishReason.TokenLimit` instead.
     */
   final case class FinishNow(cause: FinishReason, instruction: String) extends LoopDecision
 }
--- a/core/src/main/scala/sttp/ai/core/agent/AgentInterceptor.scala
+++ b/core/src/main/scala/sttp/ai/core/agent/AgentInterceptor.scala
@@ -1,0 +1,93 @@
+package sttp.ai.core.agent
+
+/** Middleware for the agent loop. Interceptors wrap iterations, LLM calls, and tool executions (onion-style, like sttp backend wrappers),
+  * and can steer the loop via [[decide]].
+  *
+  * Contracts:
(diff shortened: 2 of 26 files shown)


adamw (commented):
Ran the build locally on this branch: core tests pass on 2.13 and 3 (110 each), openai/claude/gemini tests pass on both rows (209/96/94, integration ones cancelled — no keys), `scalafmtCheckAll` and `compileDocumentation` green.

The design looks right — onion composition like sttp backend wrappers, by-name `next` for `Identity`, and the hooks adapted innermost so their behaviour doesn't change. New fields are all defaulted, so external `AgentBackend` implementations keep compiling. Tests cover the tricky cases (onion order, iteration-0 `FinishNow`, `MaxIterations` precedence, hook interleaving, spurious tool calls on the forced-final call).

Notes below, roughly in order of importance.

## 1. Six new compiler warnings on 2.13

Core on master compiles with no warnings; this branch adds six:

- `parameter ctx in method aroundIteration/aroundLlmCall/aroundToolCall is never used` — `AgentInterceptor.scala:18,21,24`. The params belong in the API, so `@nowarn("cat=unused")` on the three defaults (or a `-Wconf` entry) fixes it.
- `The outer reference in this type test cannot be checked at run time` ×2 — `Agent.scala:39,45`. `IterationOutcome` and its cases are inner classes of `Agent`, so the match on them is unchecked. Moving them out (top-level `private[agent] sealed trait IterationOutcome[F[_]]`, or into `object Agent`) removes the warning and makes exhaustiveness real.
- `value beforeToolCall in class AgentConfig is deprecated` at `AgentConfig.scala:5`. Deprecating the case class field makes the generated members warn at the definition site. Either deprecate only the builder methods, or `@nowarn` the class.

## 2. `FinishNow` takes any `FinishReason`

`LoopDecision.FinishNow(cause: FinishReason, …)` lets an interceptor report `NaturalStop`, `TokenLimit` or `Error(msg)`. `AgentInterceptorComposeSpec` does exactly that with `NaturalStop`. Since `runAs` branches on the reason, a wrong cause silently changes how the answer is parsed and misreports why the run stopped. Worth narrowing — a dedicated cause type, or a single `FinishReason.StoppedByInterceptor(reason: String)`.

## 3. `IterationInfo.isLastIteration` no longer means "tools withheld"

On a `FinishNow` iteration `includeTools = false`, but the backend still gets `IterationInfo(iteration + 1, maxIterations)`, so `isLastIteration` is false. Model selection runs off that value (`modelForIteration(iterationInfo)` in each backend), and `configuration.md` documents using exactly this flag (`if info.isLastIteration then GPT5 else GPT4oMini`). So those users won't get the stronger model on the budget-forced final answer — the one call where it matters most. Either pass the forced-final flag into `IterationInfo`, or document the gap.

## 4. Builder naming differs from the existing convention

`AgentBuilder` already has `tools(Seq)` = replace and `addTool(x)` = append. This PR uses `interceptors(Seq)` = replace and `interceptor(x)` = append, which reads the other way round. `addInterceptor` would match.

## 5. Provider mappings — two gaps

- **Claude**: `cache_creation_input_tokens` lands in `inputTokens` but not in `cachedInputTokens`, so `PriceTable` prices cache writes at the plain input rate. Anthropic charges 1.25× base input for 5-minute cache writes and 2× for 1-hour ones, so a cost budget under-counts whenever caching is used. Either add a cache-write field or document the skew.
- **Gemini**: the mapping ignores `total_tool_use_tokens` and the provider's own `total_tokens`, so `TokenUsage.totalTokens` (input + output) may not match what Gemini reports for the interaction. Also, the "output includes reasoning" invariant isn't verified here — Google's docs list `total_output_tokens` and `total_thought_tokens` as separate fields without saying whether they overlap, and this repo's live fixture (`InteractionResponseSpec.scala:56`) has a `thought` step with `total_thought_tokens: 0`. One live call with a thinking model would settle it; if thoughts are excluded, `outputTokens` should be `total_output_tokens + total_thought_tokens`.

## 6. Smaller things

- `LogLevel.Warn` is never produced anywhere in `core/src/main`. Either use it (tool or LLM failure) or drop it.
- `LoggingInterceptor` logs "started"/"finished", so a failing tool call logs only "started". A `handleError` branch would make the logs useful when things fail.
- `BudgetInterceptor` with neither `maxTotalTokens` nor `maxCost` does nothing, silently. A `require` would match the fail-fast you added for `maxCost` without `priceTable`.
- Both built-ins need `new`, and `LoggingInterceptor` needs an explicit `(IdentityMonad)` in the docs example. A companion `apply` taking the monad implicitly would read better.
- The trait, contexts and `LoopDecision` are in `sttp.ai.core.agent`, the two implementations in `…agent.interceptor`. Could all live in `interceptor`.
- A budget stop before the last iteration sends a slightly contradictory prompt: the marker renders as `[Iteration 3 of 10]` and the system prompt says the run may take 10 iterations, while the injected instruction says the budget is gone. Harmless, but the instruction could mention it is the final iteration.
- `AgentInterceptorComposeSpec` has a comment about "Task 6", which refers to the internal plan rather than anything in the repo.
- The PR description has a mangled paragraph ("…gained the `BudgetEe with exhaustive matches…"). Release notes come from PR bodies, so worth fixing.
- No test covers an interceptor that skips `next` to short-circuit a stage, which the docs list as supported (the one short-circuit test throws instead). All the new tests use `Identity`, so the lazy-`F` half of the by-name design is untested.

Nothing blocks merging except the warnings. Points 2 and 3 are worth settling before this ships in 0.8.0, since both are awkward to change later.

core/src/main/scala/sttp/ai/core/agent/AgentBuilder.scala
```diff
@@ -53,10 +53,20 @@ final class AgentBuilder[F[_], M <: AIModel] private (
   ): AgentBuilder[F, M] =
     withConfig(config.copy(responseSchema = Some(ResponseSchema.derived[T](Some(description)))))
 
+  @deprecated("Use interceptor(...) with an AgentInterceptor overriding aroundToolCall", "0.8.0")
```
adamw: just drop these since it's a new minor 0.x release

core/src/main/scala/sttp/ai/core/agent/AgentInterceptor.scala
```diff
@@ -0,0 +1,95 @@
...
+/** Accumulated run state visible to [[AgentInterceptor.decide]].
+  *
+  * @param iterationsCompleted
+  *   number of fully completed iterations (0 before the first)
+  * @param llmCalls
+  *   per-call usage breakdown; enables per-model cost budgets when the run mixes models
+  */
+final case class AgentRunState(
+    iterationsCompleted: Int,
+    maxIterations: Int,
```
adamw, line 83: so ... what's maxIterations here?

PR #575, opened 2026-08-07
Feature/typed agent io

### LINKED ISSUE #559: Typed agent input/output and composition (Agent[F, In, Out])
Give agents typed inputs and outputs (building on the existing ResponseSchema[T]) so runs return F[Either[AIException, Out]] instead of stringly results, and add composition: agentA.andThen(agentB) compiles only when A’s Out matches B’s In. Embabel matches domain types at runtime during planning; compile-time-checked handoffs are the Scala-native answer. Design questions to settle first: how In renders into the initial user message, and whether intermediate conversation history propagates across a handoff.

flsh86:
## What

Replaces the stringly-typed `Agent[F]` with `Agent[F, In, Out]`:

- **Typed output.** `deriveResponseSchema[T]` / `responseSchema[T]` now fix the agent's output type at build time, and `run` returns `F[AgentResult[Either[AgentFailure, Out]]]`. The old schema-`A`-vs-`runAs[B]` mismatch is unrepresentable; `runAs` is removed.
- **Typed input.** `input[In]` renders the input as JSON into the initial user message via its circe `Encoder` (fixed envelope); `inputRenderer[In](render: In => String)` gives explicit control.
- **Composition.** `a.andThen(b)` compiles only when A's `Out` is B's `In` (compile-time-checked handoffs — vs. Embabel's runtime type matching). `map`/`contramap` adapt almost-matching types.
- Providers (OpenAI/Claude/Gemini), `agent-testkit`, MCP tests, and all mdoc docs migrated. New cost-minimal OpenAI integration test does a real two-agent handoff.

## Key design decisions

- **Fresh conversation per handoff.** In `a.andThen(b)`, agent B starts a new conversation from A's typed `Out` value, rendered by B's own input renderer — no history carryover. The contract between agents is the type, not ambient chat state: predictable token costs, no cross-agent prompt pollution, and B's system prompt/tools apply cleanly.
- **Metadata aggregation on composition.** When both stages run: `iterations` and `usage` are summed, `toolCalls`/`llmCalls` concatenated in stage order, `finishReason` taken from the last stage that ran. On a `Left` from stage A, B never runs and A's metadata is returned as-is. Interceptors (and budgets) stay stage-local. `ToolCallRecord.iteration` is stage-local (documented); a per-stage breakdown can come later if needed.

## ⚠️ Breaking changes (target 0.9.0)

- `Agent[F]` → `Agent[F, In, Out]` (trait); `AgentBuilder[F, M]` → `AgentBuilder[F, M, In, Out]`. A fresh builder is `AgentBuilder[F, M, String, String]`, so provider entry points are source-compatible at call sites.
- `runAs[T]` removed — build with `.deriveResponseSchema[T]` (usually already present) and call `run`.
- `run` result shape: `AgentResult[String]` → `AgentResult[Either[AgentFailure, Out]]`.
- Wildcard `responseSchema(schema: ResponseSchema[_])` removed in favor of the typed overload.
- **Runtime behavior change for untyped (string) agents:** a `TokenLimit`/`Error` finish now yields `Left(AgentIncomplete(rawAnswer, reason, parseError = None))` where 0.8.0 silently returned the truncated string. `MaxIterations` and interceptor-forced stops still parse-attempt (answer may be `Right` — check `finishReason`).

## Notes

- `examples/` are kept **dual-compatible**: they compile against the released 0.5.6 artifact (CI's scala-cli check) *and* the local new API (sbt build). `TypedAgentLoopExample`'s typed pattern-match demo is a commented TODO — swap in and bump `//> using dep` after 0.9.0 is published.
- `generated-docs/out` is regenerated by the release process; intentionally untouched.
- Verified: full cross-version compile (2.13.18 / 3.3.8), 1300 JVM tests green, `compileDocumentation`, scalafmt checks.

### DIFF
--- a/agent-testkit/src/main/scala/sttp/ai/core/agent/testing/ScriptedAgent.scala
+++ b/agent-testkit/src/main/scala/sttp/ai/core/agent/testing/ScriptedAgent.scala
@@ -13,6 +13,7 @@ import sttp.shared.Identity
   * )
   * val agent = script.builder.tools(calculatorTool).build
   * val result = agent.run("What is 1 + 2?")(SyncBackendStub)
+  * // result.finalAnswer: Either[AgentFailure, String]
   * script.requests // one RecordedRequest per model round-trip
   * }}}
   *
@@ -26,7 +27,7 @@ final class ScriptedAgent[F[_]] private (script: Seq[AgentResponse])(implicit mo
   private var backends: Vector[ScriptedAgentBackend[F]] = Vector.empty
 
   /** A standard [[AgentBuilder]] wired to this script — a drop-in replacement for e.g. `OpenAIAgent.builder(...)`. */
-  def builder: AgentBuilder[F, ScriptedModel.type] = AgentBuilder[F, ScriptedModel.type] { config =>
+  def builder: AgentBuilder[F, ScriptedModel.type, String, String] = AgentBuilder[F, ScriptedModel.type] { config =>
     val backend = new ScriptedAgentBackend[F](script, config.userTools, config.systemPrompt, config.responseSchema)
     synchronized {
       backends = backends :+ backend
--- a/agent-testkit/src/test/scala/sttp/ai/core/agent/testing/AgentMatchersSpec.scala
+++ b/agent-testkit/src/test/scala/sttp/ai/core/agent/testing/AgentMatchersSpec.scala
@@ -5,7 +5,7 @@ import io.circe.generic.semiauto.deriveCodec
 import org.scalatest.exceptions.TestFailedException
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
-import sttp.ai.core.agent.{AgentResult, AgentTool, FinishReason}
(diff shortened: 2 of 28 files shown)


adamw:
Reviewed — LGTM overall. `LoopAgent` is a verbatim move of the old loop, the typed layer on top is thin and well-tested, and there are no `runAs` leftovers outside `generated-docs`.

Minor comments, none blocking:

1. After `a.andThen(b)`, a `Left` doesn't say which stage failed — a `TokenLimit` from A looks the same as one from B. Fine to fold into the deferred per-stage breakdown.
2. Consider variance now (`Agent[F, -In, +Out]`, needs `AgentResult[+T]`): it would let subtype-compatible agents compose without `map`/`contramap`, and adding it later is another breaking change.
3. `AgentBuilder[F, M, String, String]` is spelled out in ~40 provider signatures — a type alias (e.g. `AgentBuilder.Fresh[F, M]`) would cut the noise.
4. Interceptors being stage-local (so a budget on each stage is per-stage, not whole-chain) is documented in the `andThen` scaladoc, but users will miss it there — worth a sentence in the "Composing agents" section of `docs/agents/tools.md`.

PR #587, opened 2026-08-11
Add Scala 2.12 cross-building for all Scala 2 modules

### LINKED ISSUE #586: Would you accept Scala 2.12 cross-building for the Scala 2 modules?
We'd like to use sttp-ai from a large Scala 2.12 codebase where the 2.13 migration is a multi-month project in its own right. Looking into what 2.12 support would take, the gap turned out to be small: every dependency the build pins (sttp client4 4.0.26, tapir-apispec-docs 1.13.31, circe 0.14.16, circe-generic-extras 0.14.4, sttp-apispec 0.11.10, pekko 1.6.0, akka 2.6.20, scalatest 3.2.20) already publishes `_2.12` artifacts at those exact versions.

We have a working branch: 2.12.20 added to the `scala2` matrix row (core, openai, claude, gemini, agent-testkit, fs2/zio/pekko/akka streaming), ~14 small cross-compatible source adjustments, no change for 2.13/3 users, full test suite green on all three versions. The port also surfaced one latent bug worth having regardless: `Attachment` binds the wrong `Tool` type on 2.12 due to [scala/bug#4695](https://github.com/scala/bug/issues/4695) import precedence.

PR to follow shortly with an ADR documenting the compat mechanisms and an explicit exit criterion (first pinned dependency to drop `_2.12` removes the row in the next breaking release, rather than holding updates back).

Completely understand if an extra Scala version isn't a maintenance commitment you want to carry — in that case we'll maintain a fork and this issue can simply be closed.

jonoabroad:
## What

Adds Scala 2.12.20 to the build matrix for every module that already builds for 2.13: `core`, `openai`, `claude`, `gemini`, `agent-testkit`, and the fs2/zio/pekko/akka streaming modules. The Scala 3-only modules (`mcp`, `ox`, `examples`, `docs`) and Scala Native targets are unchanged. Closes #586.

Shared sources stay cross-compatible rather than duplicated (~14 small adjustments):

- 2.13-only stdlib calls replaced (`toLongOption`, `Either.orElse`, `distinctBy`).
- `sttp.ai.core.compat.unused`: `scala.annotation.unused` doesn't exist on 2.12, so it's aliased on 2.13/3 and inert on 2.12 (warning category silenced on 2.12 rows only — 2.13 keeps full checking).
- Immutable collections materialised where 2.12's default `collection.Seq` isn't accepted; type ascriptions where 2.12's inference fails.
- `Attachment` now references `assistants.Tool` fully qualified — on 2.12 it silently bound the same-package `Tool` instead of the imported one ([scala/bug#4695](https://github.com/scala/bug/issues/4695), fixed in 2.13).

ADR 0004 records the decision, the compat mechanisms, and the exit criterion; happy to drop or reword it if you'd rather record ADRs yourselves.

## Verification

`scalafmtCheckAll`, `compileDocumentation`, and the full `sbt test` aggregate pass; 2.13/3 unit suites are unchanged and green. Integration tests were **not** run (they need live API keys) — happy to run them against our own keys before merge if useful.

🤖 Generated with [Claude Code](https://claude.com/code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,7 +12,7 @@ sttp-ai is a Scala library providing a non-official client wrapper for OpenAI, C
 - Native Gemini (Google) API support via the Interactions API with dedicated module
 - OpenAI-compatible API support (Ollama, Grok, OpenRouter, etc.)
 - Streaming support for all major effect systems
-- Cross-platform: Scala 2.13.18 and Scala 3.3.8
+- Cross-platform: Scala 2.12.20, 2.13.18 and Scala 3.3.8
 - Agent loop tools loadable from [MCP](https://modelcontextprotocol.io) servers (`mcp` module, Scala 3 only, via [chimp](https://github.com/softwaremill/chimp)), in addition to manually defined `AgentTool`s
 
 ## Development Commands
@@ -137,7 +137,7 @@ Each streaming module (`streaming/{effect-system}/`) provides extensions for **a
   - Claude: Requires `ANTHROPIC_API_KEY`, minimal token usage
   - Gemini: Requires `GEMINI_API_KEY`, minimal token usage
   - Auto-skip if API key not set
-- **Cross-building**: sbt-projectmatrix for Scala 2.13.18 & 3.3.8
+- **Cross-building**: sbt-projectmatrix for Scala 2.12.20, 2.13.18 & 3.3.8
 
 ## Client Implementation Patterns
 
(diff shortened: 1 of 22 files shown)


flsh86:
Seems after newest addition to sttp-ai, it doesn't compile. Please rebase and check those:
1. ResponseSchema.oneOf (core main code) — 2.12 can't infer U through the existential Variant[_ <: U] varargs with +:. Fix: oneOfImpl[U](Seq[Variant[_ <: U]](first) ++ rest, ...).
2. A 2.12 compiler bug in the new specs — the missing-interpolator lint (force-enabled by commonSmlBuildSettings) inspects $defs/$ref inside plain string literals, and when a local val shares the name (val defs = ...downField("$defs")) it errors with recursive value defs needs type. Notably, appending -Xlint:-missing-interpolator does not override it — the flag must be filterNot-ed out of scalacOptions on 2.12 rows.

flsh86:
Closing this PR
Base and related fix in: https://github.com/softwaremill/sttp-ai/pull/605

PR #595, opened 2026-08-17
Feature/derive agent tools

### LINKED ISSUE #557: Derive tool sets from service traits (AgentTools.derive[MyService])
AgentTool.fromFunction covers single functions; deriving a whole tool set from a trait’s methods (names, scaladoc-or-annotation descriptions, parameter case classes → tapir schemas) via Scala 3 inline/Mirror machinery would match Embabel’s @Tool-on-domain-objects ergonomics without reflection. Out of scope for Scala 2.13 (document the limitation). Acceptance: a trait with two methods produces two AgentTools whose schemas round-trip through the existing AgentToolSpec machinery.

flsh86:
## AgentTools.derive: tool sets from service traits (Scala 3)

Adds `AgentTools.derive[S]` / `AgentTools.deriveF[F, S]` — a Scala 3 macro deriving a whole `Seq[AgentTool]` from a service trait's public methods, without an input case class per tool:

```scala
import sttp.tapir.Schema.annotations.description

trait WeatherService:
  @description("Get the current weather for a city")
  def currentWeather(@description("City name") city: String): String

  @description("Get a multi-day forecast")
  def forecast(city: String, days: Int): String

val tools = AgentTools.derive[WeatherService](impl) // two AgentTools

How it works

- One tool per public method with a single (possibly empty) parameter list; inherited trait methods included; accessors/vars skipped (a @description-annotated parameterless def is a compile error, not a silent skip); tools sorted by name. Case-class parameters render into $defs (named root schema).
- Each tool's input type is the tuple of the method's parameter types; the macro composes summoned per-parameter tapir Schemas and circe Decoder/Encoders, then delegates to the existing AgentTool.fromFunctionF — schema rendering goes through the same TapirSchemaToJsonSchema path as hand-built tools.
- @description (tapir's annotation) is required on methods (compile error otherwise, with fallback through overridden symbols so derive(impl) with an inferred impl-class type works), optional on parameters. Option parameters become non-required properties.
- Invalid shapes are compile errors: overloads, default parameter values, multiple/using parameter lists, method type parameters, by-name/vararg parameters, wrong return type, missing given instances. @description accepts any compile-time constant string.
- deriveF[F, S] for effectful services (F[String]-returning methods; covariantly narrowed return types conform).
- Includes a small strict-mode fix in the openai module: object roots with type but no properties (zero-arg tools) now get properties: {} added, as OpenAI strict mode requires.

Scope / compatibility

- Scala 3 only (documented); Scala 2.13 keeps using AgentTool.fromFunction. Additive, non-breaking.
- 28 new tests (AgentToolsDeriveSpec, AgentToolsDeriveErrorsSpec incl. 12 compile-error cases via typeCheckErrors, plus a strict-mode SchemaSupportSpec case); docs section added to docs/agents/tools.md (mdoc-compiled).

### DIFF
--- a/claude/src/test/scala/sttp/ai/claude/agent/ClaudeAgentBackendSpec.scala
+++ b/claude/src/test/scala/sttp/ai/claude/agent/ClaudeAgentBackendSpec.scala
@@ -59,7 +59,7 @@ class ClaudeAgentBackendSpec extends AnyFlatSpec with Matchers with EitherValues
 
     backend.convertedTools.head match {
       case raw: Tool.CustomRaw =>
-        raw.inputSchema shouldBe parse("""{"type":"object"}""").value
+        raw.inputSchema shouldBe parse("""{"type":"object","properties":{}}""").value
       case other => fail(s"expected Tool.CustomRaw, got $other")
     }
   }
@@ -101,7 +101,7 @@ class ClaudeAgentBackendSpec extends AnyFlatSpec with Matchers with EitherValues
     val backend = new ClaudeAgentBackend[Identity](client, _ => ClaudeModel.ClaudeHaiku4_5, Seq(tool), None, None)(IdentityMonad)
 
     backend.convertedTools.head match {
-      case raw: Tool.CustomRaw => raw.inputSchema shouldBe parse("""{"type":"object"}""").value
+      case raw: Tool.CustomRaw => raw.inputSchema shouldBe parse("""{"type":"object","properties":{}}""").value
       case other               => fail(s"expected Tool.CustomRaw, got $other")
     }
   }
--- a/core/src/main/scala-3/sttp/ai/core/agent/AgentTools.scala
+++ b/core/src/main/scala-3/sttp/ai/core/agent/AgentTools.scala
@@ -45,12 +45,26 @@ object AgentTools {
 
     def fail(msg: String): Nothing = report.errorAndAbort(s"AgentTools.derive[${renderType(sTpe)}]: $msg")
 
-    def descriptionOf(sym: Symbol): Option[String] =
+    def annotationTextOf(sym: Symbol): Option[String] =
       sym.getAnnotation(descriptionSym).map {
         case Apply(_, List(Literal(StringConstant(text))))              => text
         case Apply(_, List(NamedArg(_, Literal(StringConstant(text))))) => text
         case _ => fail(s"the @description annotation on '${sym.name}' must be given a string literal")
       }
+
+    // Annotations are not inherited onto overriding symbols in Scala 3, so when `S` is inferred as an implementation
+    // class (e.g. `AgentTools.derive(new WeatherImpl)`), `m` is the impl's method and carries no annotations of its
+    // own even though the trait method it overrides does. Fall back through `allOverriddenSymbols`, first hit wins.
+    def descriptionOf(sym: Symbol): Option[String] =
+      (sym :: sym.allOverriddenSymbols.toList).view.flatMap(annotationTextOf).headOption
+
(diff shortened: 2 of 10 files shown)


adamw:
Review findings. The derivation macro itself checks out (nullary methods, arities 0/1/>22, annotation inheritance, null-vs-absent encoding). The problems cluster in `SchemaSupport.ensureStrictObjectRoot` and in two derivation-semantics gaps.

**Bugs**

1. `SchemaSupport.scala:168` — the new case stamps `properties:{}` + `additionalProperties:false` onto property-less discriminated-union roots (e.g. `{"type":"object","discriminator":{...},"oneOf":[...]}`), undoing the folder's discriminator skip — no `oneOf` variant's fields validate. The PR's discriminator test uses a root *with* `properties`, which masks this.
2. `AgentTools.scala:150` — optionality mismatch for collection parameters: tapir marks `List`/`Seq`/`Map` schemas `isOptional=true`, so the emitted schema omits them from `required`, but the generated decoder uses circe's `Decoder[List[A]]`, which fails on a missing key. A model that legitimately omits the param per the schema gets a decode error instead of `Nil`. (Verified against the repo's tapir version.)
3. `SchemaSupport.scala:168` — the fix is root-only, but the invariant (per the PR's own comment) is per-object: a zero-field case-class parameter renders in `$defs` as `{"title":...,"type":"object"}` with no `properties` and OpenAI strict mode still rejects it. The folder's `onObject` is the right place.
4. `AgentTools.scala:186` — the generated decoder never checks that the argument object's keys are a subset of the parameter names, so an all-`Option` tool called with only misspelled keys (`{"limt": 5}`) decodes to all-None and executes instead of returning a decode error the loop could feed back. Claude/Gemini schemas carry no `additionalProperties:false`, so nothing masks it there.
5. `AgentTools.scala:171` — the empty-schema fix is OpenAI-only: a no-arg derived tool's schema (`{"type":"object"}` with no `properties`) is sent verbatim by the Claude/Gemini backends (`ensureObjectType` is a no-op on this shape); Gemini has rejected such schemas. No test covers a no-arg derived tool on Claude/Gemini.
6. `SchemaSupport.scala:159` — degenerate-schema normalization now lives in three non-composing variants (core `ensureObjectType` + two OpenAI root cases), and the OpenAI path still misses the boolean-schema shape core handles: an MCP tool with input schema `true` becomes `"parameters": {}` with `strict:true`, which OpenAI rejects. One canonical normalization in core would remove this class of bug.
7. `SchemaSupport.scala:168` — for `Map`/free-form object roots the new case turns a previously loud 400 into silent narrowing: the folder clobbers `additionalProperties` to `false` (pre-existing), the new case adds `properties:{}`, and strict mode now accepts the schema — the model can only ever return `{}`. Affects all `normalizeForStrict` callers, not just agent tools.
8. `SchemaSupport.scala:161` — pre-existing in the touched function: the first case clobbers `$ref`-only/`anyOf`-only roots by adding `type`/`properties`/`additionalProperties` as siblings, silently constraining the schema to `{}`. Worth guarding both cases against combinator/`$ref` roots while here.

**Cleanups**

9. `SchemaSupport.scala:168` — the two `ensureStrictObjectRoot` cases have duplicate bodies; after fixing the guards (1), fold them into one case (re-adding `type:"object"` is a no-op when present).
10. `AgentToolsDeriveSpec.scala:107` — vacuous assertion: the no-arg-tool test checks `ensureObjectType(rawJsonSchema)` has type `object`, but `ensureObjectType` inserts that key itself, so the assertion passes regardless of what the macro emits. Asserting on `rawJsonSchema` directly would also have caught 3 and 5.

PR #598, opened 2026-08-18
feat: Adding support for streaming "create responses" response

wookievx:
Brief summary of changes:

1. Introducing explicit modeling of streaming `/response` model, as defined by: https://developers.openai.com/api/docs/guides/streaming-responses.
2. Supporting fs2/zio/akka/ox.
3. Handling observed discrepancy from the official documentation: https://developers.openai.com/api/reference/resources/beta/subresources/responses/streaming-events#response.function_call_arguments.done (no name in actual responses, removed from the Scala model)
4. Added newest generation of models (gpt-5.6-x, to documented model list)


kovacshuni (approved):

flsh86 (approved):
LGTM

adamw:
Review findings:

**Bugs**

1. `ChatRequestBody.scala:360` — `ChatCompletionModel.GPT56Terra` has wire value `"gpt-5.4-terra"` instead of `"gpt-5.6-terra"` (the `ResponsesModel` counterpart is correct).
2. `ChatRequestBody.scala:361` — `ChatCompletionModel.GPT56Sol` has wire value `"gpt-5-6-sol"` instead of `"gpt-5.6-sol"`.
3. `ResponsesResponseBody.scala:520` — unknown content-part types are not covered by the forward-compat fallback: `response.content_part.added/.done` are known event types, so an unmodelled part type (e.g. a future `output_audio`) fails decoding and errors the whole stream — the case the `Unknown` event/item fallbacks were added to prevent.
4. `ChatRequestBody.scala:354` — none of the 8 new `ChatCompletionModel` case objects are in `ChatCompletionModel.values`, so e.g. `"gpt-5.2"` in a response decodes to `CustomChatCompletionModel("gpt-5.2")` and never matches the constant. (These chat-model additions also bypass the `model_update_scripts` workflow — likely how the two typos above got in.)
5. `ResponsesModel.scala:102` — `GPT56Terra` and `GPT56Sol` are missing from `ResponsesModel.values` (only `GPT56Luna` was added), so they never decode from responses.

**Cleanups**

6. `ResponsesStreamEvent.scala:450` — nothing ties the sealed-trait cases to the hand-maintained `KnownTypes` sets. Adding a new event but forgetting the `KnownTypes` entry compiles and passes tests, yet the event always decodes as `Unknown` in production. A wire-type → leaf-decoder `Map` (with `KnownTypes := map.keySet`), or a test enumerating the ADT cases against the set, would close this.
7. `ResponsesStreamEventFixture.scala:64` — `jsonCompletedWithFullUsage` splices usage into `jsonResponseBasic` by substring surgery, producing two top-level `"usage"` keys; the test passes only because circe keeps the last one. `deepMerge` on parsed JSON would be robust.
8. `ResponsesStreamEvent.scala:649` — `ResponsesStreamEvent.LogProb/TopLogProb` duplicate `OutputContent.LogProb/TopLogProb`; the same wire object surfaces as two Scala types depending on which event delivered it. `OutputTextDelta/Done` could reuse `OutputContent.LogProb`.
9. `OpenAIManualCodecs.scala:301` — `responsesStreamEventDispatch` rewrites the JSON object and re-dispatches through the derived sum decoder on every event, including every `output_text.delta`. A precomputed wire-type → decoder `Map` would avoid the rewrite and the double dispatch (and subsume `KnownTypes`, see 6).
10. `OpenAIResponsesOxStreamingIntegrationSpec.scala:87` — the function-tool-call integration test has no `maxOutputTokens`; the sibling test caps at 16. A small cap would keep the paid call cheap.

wookievx:
Resolved the issues

PR #606, opened 2026-08-27
Feature/agent conversation history

IgorWolkov:
## Multi-turn conversations: conversation history in and out of the agent loop

Makes `sttp.ai.core.agent.Agent` usable as a chat: every run returns its full `ConversationHistory`, and a run can be seeded with a prior history plus a new user message — so the caller can load previous context, inspect the transcript, and keep talking:

```scala
val agent = ClaudeAgent.synchronous(ClaudeConfig.fromEnv, ClaudeModel.ClaudeHaiku4_5.value).build

val first  = agent.run("My name is Igor. What is 2+2?")(backend)
val second = agent.run("Multiply that by 10, and remind me of my name.", first.history)(backend)

second.history.entries // full transcript: seed + new user message + tool calls/results + final answer
```

A history can also be built or restored by hand (`ConversationHistory.empty.addUserPrompt(...).addAssistantResponse(...)`) and passed as the seed of a fresh run.

## How it works

- `AgentResult` gains `history: ConversationHistory` — the complete, replayable transcript of the run: the seed (if any), the rendered user message, assistant responses, tool calls with their results, and the final assistant answer (which the loop previously discarded). It is recorded on every exit path, including token-limit and forced-final ones.
- The history is kept valid as a seed for a follow-up run: per-request decorations (iteration markers, forced-finish instructions) are transient and not part of it, and tool calls from a forced final response are not recorded — they are never executed, and an unanswered `tool_use` would make the transcript unreplayable (Claude rejects such message sequences).
- `Agent` gains `run(in, history)` as the loop entry point; the existing `run(in)` is unchanged for callers and delegates with an empty history — behaviorally identical to before. `map`/`contramap` pass the history through; `andThen` stays stage-local: a seed history feeds only the first stage, and the combined result carries the last-run stage's history (documented in the scaladoc).
- Also adds `AgentBuilder.maxTokens(n)` — a per-LLM-call output-token cap, previously hard-coded. Maps to Claude `max_tokens` (default 4096, as before), Gemini `generation_config.max_output_tokens` (default 4096, as before), and OpenAI `max_completion_tokens` (unset by default, as before; the field reasoning models accept). Useful for chat-style loops where the 4096 default is easy to hit.

## Scope / compatibility

- Source-compatible for callers: `run(in)` keeps its signature, `AgentResult.history` and `AgentConfig.maxTokens` are appended defaulted fields, and no existing behavior changes when the new features are unused. The abstract method of `Agent` moved from `run(in)` to `run(in, history)`, which affects only direct third-party `Agent` implementors (custom integrations extend `AgentBackend`, which is unchanged).
- No new dependencies; cross-built on Scala 2.13 and 3 as before.
- Tests: 7 new cases in `AgentSpec` (transcript contents, forced-final tool-call exclusion, token-limit partial answer, seeded request shape, two-turn continuation, builder config) plus request-body assertions for `maxTokens` in all three provider backend specs. Full suite green.
- Docs: "Multi-turn conversations" section in `docs/agents/quickstart.md` and `maxTokens` in `docs/agents/configuration.md` (both mdoc-compiled).

### DIFF
--- a/claude/src/main/scala/sttp/ai/claude/agent/ClaudeAgent.scala
+++ b/claude/src/main/scala/sttp/ai/claude/agent/ClaudeAgent.scala
@@ -16,9 +16,12 @@ private[claude] class ClaudeAgentBackend[F[_]](
     modelForIteration: IterationInfo => ClaudeModel,
     val tools: Seq[AgentTool[F, _]],
     val systemPrompt: Option[String],
-    responseSchema: Option[ResponseSchema[_]]
+    responseSchema: Option[ResponseSchema[_]],
+    maxTokens: Option[Int] = None
 )(implicit monad: sttp.monad.MonadError[F])
     extends AgentBackend[F] {
+
+  private val effectiveMaxTokens: Int = maxTokens.getOrElse(ClaudeAgentBackend.DefaultMaxTokens)
 
   private[claude] val convertedTools: Seq[Tool] = tools.map(convertTool)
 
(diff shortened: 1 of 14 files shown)


flsh86 (approved):
LGTM

PR #608, opened 2026-08-28
Add support for Vector Store File Batches

flsh86:
Adds the four OpenAI vector store file batch endpoints (create, retrieve, cancel, list files) to OpenAI and OpenAISyncClient, with request/response models under requests/vectorstore/file/batch and codecs for Scala 2 and 3.

Also fixes two pre-existing bugs found while validating against the live API:
- the `filter` query parameter on list requests was sent as "Completed" instead of the API's snake_case "completed" (FileStatus.toApiValue)
- ListVectorStoreFilesResponse / ListVectorStoresResponse could not decode an empty page, where the API returns first_id/last_id as null; both fields are now Option[String]

### DIFF
--- a/openai/src/main/scala-2/sttp/ai/openai/json/OpenAIDerivedCodecs.scala
+++ b/openai/src/main/scala-2/sttp/ai/openai/json/OpenAIDerivedCodecs.scala
@@ -395,8 +395,6 @@ object OpenAIDerivedCodecs {
   implicit val deleteVectorStoreFileResponseCodec: Codec[VectorStoreFileResponseData.DeleteVectorStoreFileResponse] = deriveConfiguredCodec
   implicit val createVectorStoreFileBatchBodyEncoder: Encoder[VectorStoreFileBatchRequestBody.CreateVectorStoreFileBatchBody] =
     deriveConfiguredEncoder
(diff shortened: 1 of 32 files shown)


adamw:
Reviewed against the OpenAI OpenAPI spec. Compiles and all vector store tests pass on 2.13 and 3.

The endpoints look right: paths, the `OpenAI-Beta: assistants=v2` header and the query params on list-files-in-batch all match the spec. The `filter` fix is a real bug fix - the API expects `in_progress|completed|failed|cancelled`, and `_.toString` was sending `InProgress`.

Three smaller things:

1. `VectorStoreFileBatchResponseData.scala:12` - the scaladoc says the object type is always `vector_store.file_batch`, and both fixtures in `VectorStoreFileBatchFixture` use that value. The API actually returns `vector_store.files_batch` (plural); the spec's enum is plural, only its text description has the singular typo. Decoding is unaffected since it's a plain `String`, but the docs and fixtures don't match real responses.

2. The `null` cursor fix on empty pages is only applied to the two vector store list responses. The same `firstId: String, lastId: String` shape is still in `ListMessagesResponse`, `ListRunsResponse`, `ListRunStepsResponse`, `InputItemsListResponseBody`, `ListMessageResponse`, `ListChatResponse` and `ListAssistantsResponse` - listing an empty thread or an empty assistant list fails to decode the same way.

3. Changing `firstId`/`lastId` from `String` to `Option[String]` breaks source and binary compatibility, and there's no MiMa here to catch it. Worth a note in the release notes.

